PR #3, opened 2022-08-03
User readable printing of observation and action spaces

StoneT2000:


Jiayuan-Gu (commented):
The verbose print is due to gym>0.18.3. It seems to be fixed in gym>0.21.0, but the higher version has other issues. Thus, I suggest `pip install gym==0.18.3 --no-deps` in README, and it is also the default in the docker.

PR #4, opened 2022-08-04
Improved functionality of a few utility scripts

xiqiangliu:


tools/replay_trajectory.py
```diff
@@ -407,6 +413,13 @@ def main():
     args = parse_args()
 
     if args.num_procs > 1:
+        if args.verbose:
+            tqdm.write("Checking environment type (rigid/soft-body)")
+
+        if detect_mpm(args.traj_path):
```
Jiayuan-Gu: Can `spawn` also be applied to envs other than MPM ones?

xiqiangliu: They can but it's not necessary. Using `spawn` mode would repeatedly print logs when importing `sapien` which is less than ideal.

Jiayuan-Gu (approved):

PR #9, opened 2022-08-09
Update README.md

### LINKED ISSUE #8: [Minor] Default observation in some soft body demos seems to be `pointcloud` (leads to different expected behavior).
The current README states for `tools/replay_trajectory.py`:

> We recommend using our script only for converting actions into different control modes and not record any observation information (i.e. without passing `--obs-mode`). The reason is that (1) some observation modes, e.g. point cloud, can take much space without any post-processing, e.g., point cloud downsampling; in addition, the state mode for soft-body environments also has a similar issue, since the states of those environments are particles. (2) Some algorithms (e.g. GAIL) require custom keys stored in the demonstration files, e.g. next-observation.

However it seems like for 3 of the soft body envs, after loading the demonstrations, the `tools/replay_trajectory.py` will save using the point cloud type even without `--obs-mode` explicitly passed on the command line. Thus, by default the code will do the "non recommended" behavior as per the README. 

Running this command:

```
python tools/replay_trajectory.py --traj-path demos/soft_body_envs/Excavate-v0/trajectory.h5   --save-traj --target-control-mode pd_joint_delta_pos --num-procs 10
```

Will create:

```
demos/soft_body_envs/Excavate-v0:
total 21G
-rw-rw-r-- 1 seita seita 219M Aug  4 11:28 trajectory.h5
-rw-rw-r-- 1 seita seita  57K Aug  4 11:28 trajectory.json
-rw-rw-r-- 1 seita seita  21G Aug  9 10:59 trajectory.pointcloud.pd_joint_delta_pos.h5
-rw-rw-r-- 1 seita seita  60K Aug  9 10:59 trajectory.pointcloud.pd_joint_delta_pos.json
```

Whereas for another env like `PegInsertionSide-v0` I will see:

```
demos/rigid_body_envs/PegInsertionSide-v0:
total 97M
-rw-rw-r-- 1 seita seita  49M Aug  4 11:27 trajectory.h5
-rw-rw-r-- 1 seita seita 398K Aug  4 11:27 trajectory.json
-rw-rw-r-- 1 seita seita  48M Aug  9 10:23 trajectory.none.pd_joint_delta_pos.h5
-rw-rw-r-- 1 seita seita 403K Aug  9 10:23 trajectory.none.pd_joint_delta_pos.json
```

So here we have `trajectory.none` as the prefix instead of `trajectory.pointcloud` and the file size is much smaller for the former.

The reason is I think that the `trajectory.json` files we load has `pointcloud` as the `obs_mode` for soft body envs, for example:

```
(mani_skill2) seita@takeshi:~/ManiSkill2 (main)$ head -n 20 demos/soft_body_envs/Excavate-v0/trajectory.json 
{
  "env_info": {
    "env_id": "Excavate-v0",
    "max_episode_steps": 250,
    "env_kwargs": {
      "obs_mode": "pointcloud",
      "reward_mode": "dense",
      "control_mode": "pd_joint_pos"
    }
  },
  "commit_info": {
    "commit_id": "ef73ec44f5f0f22c9b2f71e682fbdca0c7d377fa",
    "branch": "yihe-dev"
  },
  "episodes": [
    {
      "episode_id": 0,
      "episode_seed": 1,
      "reset_kwargs": {
        "seed": 1,
(mani_skill2) seita@takeshi:~/ManiSkill2 (main)$ 
```

The same is true with `Fill-v0` and `Hang-v0` though interestingly not `Pour-v0` (which uses `particle` for some reason). The rigid body envs have `none` as the observation mode.

To be clear, this doesn't affect the _correctness_ of the code, though it might be easier (possibly for future users?) if they can use the "none" observation type to save on disk space and for ease of compatibility with the ManiSkill2-Learn commands which seem to expect `trajectory.none`. Or maybe a possible remark to this in the README could help clarify?

Thanks for the code and environments. :)

xuanlinli17:


README.md
```diff
@@ -179,13 +186,14 @@ python tools/replay_trajectory.py --traj-path demos/rigid_body_envs/PickCube-v0/
 
 - `--save-traj`: save the replayed trajectory to the same folder as the original trajectory file.
 - `--num-procs=10`: split trajectories to multiple processes (e.g., 10 processes) for acceleration.
+- `--obs-mode=none`: (not included in the script above) specify the observation mode as `none`, i.e. not saving any observations.
```
DanielTakeshi: Minor: can remove "(not included in the script above)"

README.md
```diff
@@ -168,7 +168,14 @@ The raw demonstration files contain all necessary information (e.g. initial stat
 ```bash
 # Replay demonstrations with control_mode=pd_joint_delta_pos
 python tools/replay_trajectory.py --traj-path demos/rigid_body_envs/PickCube-v0/trajectory.h5 \
-  --save-traj --target-control-mode pd_joint_delta_pos --num-procs 10
+  --save-traj --target-control-mode pd_joint_delta_pos --obs-mode none --num-procs 10
+  
+# For soft-body environments, if you haven't used an environment in the past and
```
Jiayuan-Gu: Please move the comment somewhere. For example, adding a ">" (markdown quote) paragraph below the code block.

Maybe we can remove the toggle list for warp installation. Fanbo has already provided a script to precompile SDF for soft-body environments, and we can link that.

Jiayuan-Gu (approved):

PR #11, opened 2022-08-15
enable shadow by default in examples/demo_manual_control.py

xuanlinli17:


Jiayuan-Gu:
The user may expect the same visual result during training if the shadow is enabled by default.

PR #31, opened 2022-10-05
Fixed edge case handling in RecordEpisode wrapper

xiqiangliu:
Due to empty action history, the current `RecordEpisode` wrapper will throw out an error when a user resets twice under the default setting, as `_episode_data` is not empty. The same would also happen if `flush_trajectory()` is manually called right after the reset when `save_on_reset` is turned off.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ python examples/demo_manual_control.py -e PickCube-v0
 # PickCube-v0 can be replaced with other environment id.
 ```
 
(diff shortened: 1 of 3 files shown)


Jiayuan-Gu:
It seems to be an expected behavior, since the observation after resetting the environment is a part of the trajectory. Can you elaborate on the cases that cause inconvenience when you want to ignore observations after reset?

Jiayuan-Gu (approved):

PR #32, opened 2022-10-30
Add option to enable recording option in evaluation mode

fabid:
This PR enable the recording of evaluation videos via the `--record-dir`.
The option lets the user specify a directory to record simulation videos. If no directiory is specified, no video is recorded, as is currently the case.
This is useful when doing evaluations to visualize and debug fail attempts


Jiayuan-Gu:
Thank you for the pull request. We have refactored the evaluation code  and implemented your request at #38

PR #44, opened 2023-01-06
Refactor for vectorized environment supported by RenderServer

Jiayuan-Gu:
To test this pull-request, please first install the nightly-built SAPIEN. Please refer to `examples/demo_vec_env.py` for the usage of `VecEnv`.

Here are important modifications for reference:
- Introduce `CameraConfig` to configure cameras to know demanded resources before creating actual `sapien.CameraEntity`
- All the raw environments only support `image` for visual observations. `rgbd` and `pointcloud` are supported by extra wrappers. But we make efforts to keep users unaffected if they use `gym.make` to create environments.
- use `enable_robot_seg` to enable robot segmentation instead of specifying observation modes like `rgbd_robot_seg`
- introduce `VecEnv` to make use of `sapien.RenderServer` to accelerate visual observations


examples/demo_vec_env.py
```diff
@@ -0,0 +1,118 @@
...
+            gym.make,
+            id=args.env_id,
+            obs_mode=args.obs_mode,
+            reward_mode=args.reward_mode,
+            control_mode=args.control_mode,
+            **args.env_kwargs,
+        )
+        env_fns.append(env_fn)
+
+    # env = VecEnv(env_fns, server_address="localhost:15003")
```
StoneT2000: what's this for? I assume it's just to show another way of using VecEnv instead of using make. Maybe add comment about the other way. It's a little unclear that the previous env_fns list is actually not relevant for the uncommented code

mani_skill2/agents/base_agent.py
```diff
@@ -81,7 +83,7 @@ def __init__(
         self._load_articulation()
         self._setup_controllers()
         self.set_control_mode(control_mode)
-        self._setup_cameras()
+        # self._setup_cameras()
```
StoneT2000: Can we just remove this line instead of commenting out?

StoneT2000 (changes requested):

examples/demo_vec_env.py
Jiayuan-Gu: Yes, they are some legacy codes and I should either comment them as well or just remove them in a clean-up commit.

mani_skill2/agents/base_agent.py
Jiayuan-Gu: Will do in a clean-up commit

PR #45, opened 2023-01-07
Add python api to download demonstrations and fix gdown bug for large file downloads

StoneT2000:


StoneT2000:
This also adds a fix for gdown from https://github.com/wkentaro/gdown/issues/43 where large files couldn't be downloaded and it would incorrectly say the error is because there's no permissions.

.gitignore
```diff
@@ -33,3 +33,5 @@ imgui.ini
 # but can be excluded explicitly
 demos/*
 tmp/*
+
+examples/tutorials/demos
```
Jiayuan-Gu: I suggest you create a new .gitignore under `examples/tutorials` instead of modifying the .gitignore under the root directory.

mani_skill2/utils/dataset_download.py
```diff
@@ -0,0 +1,54 @@
...
+    download a demonstration dataset for a specific environment
+    """
+    if env_id not in DATASET_SOURCES:
+        raise ValueError(f"{env_id} is not a valid environment id with demonstrations")
+    dataset_metadata = DATASET_SOURCES[env_id]
+    output_dir = osp.join(output_dir, dataset_metadata["env_type"], env_id)
+
+    # we first try the google drive link
+    print(f"Downloading from Google Drive link {dataset_metadata['gd_url']} ...")
+    import gdown
```
Jiayuan-Gu: 1. Please use `black` and `isort` to format your python file. I can add a pre-commit hook if you want.
2. It might be better to import `gdown` at the top since it is a dependency of this module.

mani_skill2/utils/dataset_download.py
```diff
@@ -0,0 +1,54 @@
...
+DATASET_SOURCES["StackCube-v0"] = dict(env_type="rigid_body", s1_url="", gd_url="https://drive.google.com/drive/folders/1XSSsI58rpLYxyexbNFf7LzhJQXAY_fVO?usp=share_link")
+DATASET_SOURCES["PickCube-v0"] = dict(env_type="rigid_body", s1_url="", gd_url="https://drive.google.com/drive/folders/1WgYpQjqnZBbyXqlqtQfoNlCKAVPdeRIx?usp=share_link")
+# Soft body envs
+DATASET_SOURCES["Write-v0"] = dict(env_type="soft_body", s1_url="", gd_url="https://drive.google.com/drive/folders/1ziY-cL_ofq52zYapdQIJNJWnaZZPGVO9?usp=share_link")
+DATASET_SOURCES["Pinch-v0"] = dict(env_type="soft_body", s1_url="", gd_url="https://drive.google.com/drive/folders/1F6Yx1kqZ8mg2H0ShqNWnwsTEigcnAEol?usp=share_link")
+DATASET_SOURCES["Hang-v0"] = dict(env_type="soft_body", s1_url="", gd_url="https://drive.google.com/drive/folders/1zpU__pG-N7SEhIH4YD87MeeGRDfoGq5_?usp=share_link")
+DATASET_SOURCES["Pour-v0"] = dict(env_type="soft_body", s1_url="", gd_url="https://drive.google.com/drive/folders/1jlZ3K6nOXgqjXeKVRYNKdXIbTRau_1Ge?usp=share_link")
+DATASET_SOURCES["Excavate-v0"] = dict(env_type="soft_body", s1_url="", gd_url="https://drive.google.com/drive/folders/1G3eycgS1ckJlG5R5Z6wClCAPHuSg94ei?usp=share_link")
```
Jiayuan-Gu: what does `s1_url` mean?

mani_skill2/utils/dataset_download.py
```diff
@@ -0,0 +1,54 @@
...
+    print(f"Downloading from Google Drive link {dataset_metadata['gd_url']} ...")
+    import gdown
+    result = gdown.download_folder(dataset_metadata["gd_url"], output=output_dir)
+    if result == False:
+        print("Google drive link failed")
+    # TODO add mirror links
+
+
+    if result == False:
+        print("All links failed, please contact an admin for this issue")
```
Jiayuan-Gu: Should it be sth like "reporting to maintainers" or "create an issue on Github"? I think "admin" sounds strange to users.

mani_skill2/utils/dataset_download.py
```diff
@@ -0,0 +1,54 @@
+import os.path as osp
```
Jiayuan-Gu: Can you add a main function so that users can download from the command line as well

Jiayuan-Gu (changes requested):

mani_skill2/utils/dataset_download.py
StoneT2000: mirror link for storage 1 if we have one in the future

Jiayuan-Gu (approved):

PR #51, opened 2023-01-12
Make ManiSkill2 pip installable

Jiayuan-Gu:
- `mani_skill2.assets` contain necessary robot descriptions and task assets, which will be packaged into the wheel.
- Use `mani_skill2.PACKAGE_ASSET_DIR` to manage assets that will be packaged into wheels, and use `mani_skill2.ASSET_DIR` to manage assets require downloading.
- Refactor `download` to `download_asset`, and support downloading by group and environment id. For example, `python -m mani_skill2.utils.download_asset PickClutterYCB-v0`


mani_skill2/utils/download_asset.py
```diff
@@ -0,0 +1,263 @@
...
+        "uid",
+        type=str,
+        help="Asset UID. Use 'all' to download all assets.",
+    )
+    parser.add_argument("--quiet", action="store_true", help="Disable verbose output.")
+    parser.add_argument(
+        "--non-interactive", action="store_true", help="Disable prompts."
+    )
+    return parser.parse_args()
+
```
StoneT2000, line 221: Should we add something to skip re-downloading already downloaded assets? might make running things a bit easier without having to check yourself with workloads that don't spin new instances all the time.

might also want to do this with datasets altho versioning is a problem perhaps

setup.py
```diff
@@ -1,8 +1,33 @@
...
+        exclude=["warp_maniskill.warp.tests"],
+    ),
+    python_requires=">=3.7",
+    install_requires=read_requirements(),
+    # Glob patterns do not automatically match dotfiles
+    package_data={
+        "mani_skill2": ["assets/**", "envs/mpm/shader/**"],
+        "warp_maniskill.warp": ["native/*", "native/nanovdb/*"],
+    },
+    exclude_package_data={"": ["*.convex.stl"]},
```
StoneT2000, line 32: have you tried to package this to the test pypi server yet? is there smth I can test with / build with?

StoneT2000 (changes requested):

mani_skill2/utils/download_asset.py
Jiayuan-Gu: Skip re-downloading is already implemented (but it requires users to decide).

The old assets are still kept for backward compatibility. The versioning can somehow be implemented like Habitat, but may need some re-designing.

setup.py
Jiayuan-Gu: If you want to test it locally, you can just `pip install .`

PR #58, opened 2023-01-31
fix PickClutter bug where obj_start_pos is not an np array

xuanlinli17:


mani_skill2/envs/pick_and_place/pick_clutter.py
```diff
@@ -151,7 +151,7 @@ def _set_target(self):
         actor_idx = random_choice(visible_inds, self._episode_rng)
         self.obj = self.objs[actor_idx]
         obj_rep_pts = self.episode["actors"][actor_idx]["rep_pts"]
-        self.obj_start_pos = random_choice(obj_rep_pts, self._episode_rng)
+        self.obj_start_pos = np.array(random_choice(obj_rep_pts, self._episode_rng))
```
Jiayuan-Gu: Could you use `np.float32` instead?

xuanlinli17:
done

Jiayuan-Gu (approved):

PR #72, opened 2023-03-22
Fix tutorial installation

StoneT2000:


Jiayuan-Gu:
Hi @StoneT2000 , can you describe what issues are encountered, and why is a specific version of `setuptools` needed?

StoneT2000:
I think the default setup tools that comes with colab was updated recently. The current notebooks all can't install gym as a result.

This is also a sign that we should shift to the gymnasium API in the future when we can.

Jiayuan-Gu (approved):
Please squash merge this pull request. Thx.

PR #135, opened 2023-08-16
update versions to 0.5.0 and fix docs with downgrade of sphinx

StoneT2000:


Jiayuan-Gu (approved):
Please add an explanation (e.g., relevant issue in sphinx or other repos) so that we can remove this workaround when sphinx fixes the issue.

StoneT2000:
> Please add an explanation (e.g., relevant issue in sphinx or other repos) so that we can remove this workaround when sphinx fixes the issue.

added a comment to our setup.py

PR #137, opened 2023-08-17
Setup Automatic Pypi uploads on release

StoneT2000:


StoneT2000:
needs 2 github secrets filled out

TEST_PYPI_API_TOKEN and PYPI_API_TOKEN

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,38 @@
...
+        --wheel
+        --outdir dist/
+        .
+    - name: Publish distribution 📦 to Test PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository-url: https://test.pypi.org/legacy/
+    - name: Publish distribution 📦 to PyPI
+      if: startsWith(github.ref, 'refs/tags')
```
Jiayuan-Gu: Does it mean any tagged version will be released? I am not sure whether it is a good option. Is there a reference for such Github actions?

StoneT2000:
This is now configured to only publish to pypi and testpypi on tagged commits. It will only publish to pypi if the tagged commit starts with a "v", which I have configured Github to protect that tag name, so only maintainers can add release tags starting with"v".

xuanlinli17 (approved):

PR #180, opened 2023-12-18
Added decision transformer implementation to baseline

arnavg115:
I added a decision transformers implementation as part of a new baselines directory.


examples/baselines/decision-transformer/DecisionTransformer.ipynb
StoneT2000, line 1: this file along with all otherfiles besides the Dockerfile should be lowercased

examples/baselines/decision-transformer/Dockerfile
StoneT2000, line 1: we will probably try to make the image on haosulab, this is okay for now.

examples/baselines/decision-transformer/decision_transformer/Dataset.py
```diff
@@ -0,0 +1,113 @@
...
+        s[-1] = np.concatenate([np.zeros((1, max_len - tlen, state_dim)), s[-1]], axis=1)
+        s[-1] = (s[-1] - state_mean) / state_std
+        a[-1] = np.concatenate([np.ones((1, max_len - tlen, act_dim)) * -10., a[-1]], axis=1)
+        r[-1] = np.concatenate([np.zeros((1, max_len - tlen, 1)), r[-1]], axis=1)
+        d[-1] = np.concatenate([np.ones((1, max_len - tlen)) * 2, d[-1]], axis=1)
+        rtg[-1] = np.concatenate([np.zeros((1, max_len - tlen, 1)), rtg[-1]], axis=1)
+        timesteps[-1] = np.concatenate([np.zeros((1, max_len - tlen)), timesteps[-1]], axis=1)
+        mask.append(np.concatenate([np.zeros((1, max_len - tlen)), np.ones((1, tlen))], axis=1))
+
+    device = "cuda:0"
```
StoneT2000: this code would only run on pytorch and on the first gpu. I think original DT wrote it like this but it's not good practice. Can you factor out the device variable to be an function argument? And then in training when defining sequence trainer you can say which device you want to use

```
device = "cuda" if torch.cuda.is_available() else: "cpu"
```
don't remember exact strings/api but this is a pretty common line to use to auto detect which device to put data on.

examples/baselines/decision-transformer/decision_transformer/Eval.py
```diff
@@ -0,0 +1,115 @@
+import torch
+import numpy as np
+
```
StoneT2000: can you make sure to run the code formatter? 

```
pip install isort black
```

Then run

```
isort mani_skill2
black -l mani_skill2
```

to auto format the code (we should add pre-commit hooks later.)

examples/baselines/decision-transformer/decision_transformer/Eval.py
```diff
@@ -0,0 +1,115 @@
...
+        episode_return += reward
+        episode_length += 1
+
+        # if done:
+        #     break
+    success = info["success"]
+    return episode_return, episode_length, success
+
+def eval_episodes(target_rew,env, variant):
+  def fn(model):
```
StoneT2000: change all code to be 4 spaces (code formatter will probably do this for you)

StoneT2000 (changes requested):
Code looks mostly good and as discussed previously the training results are sufficient for now without need to tune too much.

Some other things is can you make a README.md file for this baselines/decision-transformer folder detailing 1. Original code repository, 2. how to setup the environment (there seems to be missing a requirements.txt file here to install dependencies) 3. How to run training 4. Post a link to weight and biases results

examples/baselines/decision-transformer/Readme.md
```diff
@@ -0,0 +1,21 @@
+# Decision Transformer
+
+This baseline is adapted from the `gym` example provided [here](https://github.com/kzl/decision-transformer/tree/master). This folder provides both a jupyter notebook version that can be run on Google Colab as well as a python script that can be used for training a model. A few environments have already been trained on and these runs can be viewed [here](https://wandb.ai/ag115115/decision-transformer).
+
```
StoneT2000, line 4: last thing, add instructions for how to install requirements (it should be obvious but we need to make peoples lives easier). Make sure to mention that they need to install maniskill2 first.

arnavg115: I added this in my most recent commit.

PR #187, opened 2024-01-22
Added implementation for VisualEncoder for visual processing

itsygao:
Added the implementation of VisualEncoder that inherits VecEnvObservationWrapper. Use pretrained ResNet18 from r3m as visual backbone, which automatically processes all RGB images (depth is discarded).

Added RL PPO code that uses VisualEncoder in `examples/tutorials/reinforcement-learning`. Achieved >90% success rate on dense rewards on LiftCube. Results available at https://wandb.ai/itsygao/VisualEncoder_results

Side note: r3m is not available on PyPI so we need to run `pip install git+https://github.com/facebookresearch/r3m.git@main` instead


StoneT2000:
thanks! Will modify code and integrate it as part of the new maniskill

PR #204, opened 2024-02-05
trivial pr: `cameras` render mode

arth-shukla:
ik we changed cameras to sensors for a reason, but is it ok to add this back since it's useful:

changes
- added camera render which combines human cameras and agent sensors
- fix inverted color bug for human cameras on cpu
- small typing stuff

justification
- it's useful to see both in my log videos
- render_sensors is too small to fit all info_on_video content (since human render is 512x512, while fetch's sensors are 128x128)

testing
- my fetch rl code

https://github.com/haosulab/ManiSkill2/assets/44527268/9ae4a84b-bdc3-4d2c-8738-e9a2b4b7b3ba

not addressed here
- currently vis utils not written in `physx.is_gpu_enabled` format; did not change since iirc we might be moving away from this. if we don't then will need to change later

### DIFF
--- a/mani_skill2/envs/sapien_env.py
+++ b/mani_skill2/envs/sapien_env.py
@@ -1009,7 +1009,7 @@ class BaseEnv(gym.Env):
                     continue
                 camera.capture()
                 # TODO (stao): the output of this is not the same as gpu setting, its float here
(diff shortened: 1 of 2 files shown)


StoneT2000 (changes requested):
just need to remove adding back cameras setting

StoneT2000 (approved):

PR #221, opened 2024-03-02
Fix actor show/hide visual gpu sim

arth-shukla:
tested on gpu and cpu sim, checked depth camera obs as well

## **REIMPLEMENTED**
- by default hidden objs are hidden
- if hidden actor pose is changed, save to `self.before_hide_pose` so it can be applied once actor is un-hidden

### DIFF
--- a/mani_skill2/envs/sapien_env.py
+++ b/mani_skill2/envs/sapien_env.py
@@ -446,15 +446,12 @@ class BaseEnv(gym.Env):
             obj.hide_visual()
         self._scene.update_render()
         self.capture_sensor_data()
-        obs = OrderedDict(
+        return OrderedDict(
             agent=self._get_obs_agent(),
             extra=self._get_obs_extra(info),
             sensor_param=self.get_sensor_params(),
             sensor_data=self.get_sensor_obs(),
         )
-        for obj in self._hidden_objects:
-            obj.show_visual()
-        return obs
 
     @property
     def robot_link_ids(self):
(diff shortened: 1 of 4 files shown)


mani_skill2/envs/sapien_env.py
```diff
@@ -969,9 +972,6 @@ def render_human(self):
                 self._viewer.set_camera_pose(
                     self._human_render_cameras["render_camera"].camera.global_pose
                 )
-
-        for obj in self._hidden_objects:
-            obj.show_visual()
```
StoneT2000, line 974: why is this removed? Ghost objects/visuals should show up in human renders

StoneT2000: oh is this because by default it is always shown except for rendering observations? I think we should do this. The default should be it is always hidden. Only when render_human is called is it shown.

StoneT2000 (changes requested):

mani_skill2/envs/sapien_env.py
StoneT2000: Well to be honest it may not matter, we don't call any of the expensive operations if the object is already hidden

StoneT2000: actually ignore this. re-read ur comments

arth-shukla:
added code to actors so if pose is updated when hidden, it changes `self.before_hide_pose` instead

arth-shukla:
added unit test

StoneT2000 (approved):

PR #241, opened 2024-03-12
Added ppo partial reset arg

arnavg115:


StoneT2000:
the same should be done for evaluation environments. Otherwise training success rates and rewards don't line up well with eval results

PR #245, opened 2024-03-15
New tasks with assets from AI2 Thor dataset

smtrived:


mani_skill/envs/tasks/egg_to_bowl.py
```diff
@@ -0,0 +1,109 @@
...
+
+            xyz_e = torch.zeros((b, 3))
+            xyz_e[..., :2] = torch.rand((b, 2)) * 0.2 - 0.1
+            xyz_e[..., 2] = 0.04
+            q = euler2quat(0, 0, 0)
+
+            obj_pose = Pose.create_from_pq(p=xyz_e, q=q)
+            self.egg.set_pose(obj_pose)
+            
+            xyz_b[..., :2] = xyz_e[..., : 2] - 0.1
```
StoneT2000, line 82: there is a runtime error here, xyz_b not defined

mani_skill/envs/tasks/egg_to_bowl.py
```diff
@@ -0,0 +1,109 @@
...
+        return CameraConfig("render_camera", pose, 512, 512, 1, 0.01, 100)
+
+    def _load_scene(self):
+        self.table_scene = TableSceneBuilder(
+            env=self, robot_init_qpos_noise=self.robot_init_qpos_noise
+        )
+        self.table_scene.build()
+
+        self.egg = actors.build_actor_ai2_helper(
+            model_id="egg_1",
```
StoneT2000: these need to be uppercased

mani_skill/utils/building/actors.py
```diff
@@ -344,3 +344,29 @@ def build_actor_ai2(
     if set_object_on_ground:
         actor.set_pose(sapien.Pose(p=[0, 0, 0]))
     return actor
+
+def build_actor_ai2_helper(
```
StoneT2000, line 348: i think we will discuss how to unify the building API in the future so we might change this function

mani_skill/envs/tasks/pour_bowl_contents.py
```diff
@@ -0,0 +1,124 @@
...
+from mani_skill.agents.robots import Fetch, Panda, Xmate3Robotiq
+from mani_skill.envs.sapien_env import BaseEnv
+from mani_skill.sensors.camera import CameraConfig
+from mani_skill.utils.building import actors
+from mani_skill.utils.geometry import rotation_conversions
+from mani_skill.utils.registration import register_env
+from mani_skill.utils.sapien_utils import look_at
+from mani_skill.utils.scene_builder.table.table_scene_builder import TableSceneBuilder
+from mani_skill.utils.structs.pose import Pose
+from mani_skill.utils.structs.types import Array
```
StoneT2000: make sure not to import unused packages (I realize some of other code does this and will fix that).

mani_skill/envs/tasks/book_to_shelf.py
```diff
@@ -0,0 +1,109 @@
...
+        return CameraConfig("render_camera", pose, 512, 512, 1, 0.01, 100)
+
+    def _load_scene(self):
+        self.table_scene = TableSceneBuilder(
+            env=self, robot_init_qpos_noise=self.robot_init_qpos_noise
+        )
+        self.table_scene.build()
+
+        self.book = actors.build_actor_ai2_helper(
+            model_id="Book_14",
```
StoneT2000, line 52: I think the task idea is nice but looks a little too strange (the shelf is longer than the table and stretches off of it).

Could we somehow make a scene where there is a tall bookshelf situated right next to the table? E.g. like a shelf against a wall and right in front of the shelf is the table and the robot is there.

mani_skill/envs/tasks/book_to_shelf.py
```diff
@@ -0,0 +1,109 @@
...
+
+            obj_pose = Pose.create_from_pq(p=xyz, q=q)
+            self.book.set_pose(obj_pose)
+            xyz[..., 1] -= 0.3
+            xyz[..., 2] += 0.15
+            q = euler2quat(0, 0, np.pi)
+            obj_pose = Pose.create_from_pq(p=xyz, q=q)
+            self.shelf.set_pose(obj_pose)
+
+    def evaluate(self):
```
StoneT2000: for figuring out where the book should go in a region of space (like an entire shelf), you likely need to hardcode a few things. Namely figure out relative to the shelf's pose, where is each region of space.

For a book shelving task you can consider doing the easy one where as long as the book is inside the book shelf (check the bounding box of the shelf with the bounding box of the book, the book should be completely inside the shelf or something).

We can make a harder version like "InsertBookToShelfLevel2-v1" or something where the evaluation criterion is the specific shelf (marked by a green ball perhaps). that is future work.

StoneT2000 (changes requested):

PR #249, opened 2024-03-22
Rotate Cube Task using Trifingerpro robot

Kami-code:
Rotate Cube Task using Trifingerpro robot


mani_skill/agents/robots/trifingerpro/trifingerpro.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+@register_agent()
+class TriFingerPro(BaseAgent):
+    """
+    Modified from https://github.com/NVIDIA-Omniverse/IsaacGymEnvs/blob/main/isaacgymenvs/tasks/trifinger.py
+
+    """
+    uid = "trifingerpro"
+    urdf_path = f"{PACKAGE_ASSET_DIR}/robots/trifinger/trifingerpro.urdf"
```
StoneT2000, line 27: it doesn't seem like these assets are in the repo? can you add them

mani_skill/envs/tasks/rotate_cube.py
```diff
@@ -0,0 +1,438 @@
+"""
```
StoneT2000: remove the comments up here. That is just for tutorial

mani_skill/envs/tasks/rotate_cube.py
```diff
@@ -0,0 +1,438 @@
...
+        total_reward = finger_reach_object_reward + pose_reward
+        total_reward[info["success"]] = 3
+        return total_reward
+
+    def compute_normalized_dense_reward(self, obs: Any, action: Array, info: Dict):
+        return self.compute_dense_reward(obs=obs, action=action, info=info) / 1
+
+
+@torch.jit.script
+def quat_from_euler_xyz(roll, pitch, yaw):
```
StoneT2000: our other code is using snippets from pytorch3d for rotation conversions. Is this a fast implementation compared to that one? If so we should move this code to the mani_skill/utils/geometry/rotation_conversions.py file

mani_skill/agents/robots/trifingerpro/trifingerpro.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+        Get the proprioceptive state of the agent.
+        """
+        obs = super().get_proprioception()
+        obs.update({"tip_poses": self.tip_poses().view(-1, 21)})
+        obs.update({"tip_velocities": self.tip_velocities().view(-1, 9)})
+        return obs
+
+    # @property
+    def tip_poses(self):
```
StoneT2000, line 163: why not make these properties?

mani_skill/envs/tasks/rotate_cube.py
```diff
@@ -0,0 +1,438 @@
...
+        )
+
+        self.obj_goal = actors.build_colorful_cube(
+            self._scene,
+            half_size=self.size / 2,
+            color=np.array([12, 42, 160, 255]) / 255,
+            name="cube_goal",
+            body_type="kinematic",
+            add_collision=False
+        )
```
StoneT2000, line 123: I recommend making goals a green color (unless you have a good reason not to use green). I think this is the blue color from some of the other tasks we had iirc.

Another thing is because this is actually not part of an image observation, this goal should be added to the hidden objects list, which are all hidden when rendering visual observations (but shown during render_mode=="rgb_array" or "human")

```
self._hidden_objects.append(self.obj_goal)
```

StoneT2000 (changes requested):
Awesome work so far. 

A few additional things that need to be done:

Can you save the script you run for ppo training to examples/baselines/ppo/readme.md (just write briefly what you run). Our baseline documentation is not super organized at the moment but in the future every task will have info about how to train on it with dense reward / demos.

Another, can you generate, with ray tracing on, a video of the robot solving each level of task? you can load the trained model and turn num_eval_envs=1 and for env_kwargs add `shader_dir="rt-med".

Then upload that video to `figures/environment_demos` similar to the ones there already.

Finally to document this task, please add an entry for it similar to the other ones I have done in docs/source/tasks/index.md. See how i did it for RotateValveLevel0-v1

mani_skill/agents/robots/trifingerpro/trifingerpro.py
Kami-code: I think I did add it. Isn't it the right place to put it? https://github.com/haosulab/ManiSkill2/blob/7c6d1d64ada6a4c06ce9a23c0a468f85350a74dd/mani_skill/assets/robots/trifinger/trifingerpro.urdf

mani_skill/envs/tasks/rotate_cube.py
Kami-code: Thanks for reminding me. I just forgot about this library.

mani_skill/agents/robots/trifingerpro/trifingerpro.py
StoneT2000: must have missed the file.

mani_skill/envs/tasks/rotate_cube.py
```diff
@@ -0,0 +1,411 @@
...
+    roll = 2 * np.pi * torch.rand(num, dtype=torch.float, device=device)
+    pitch = torch.zeros(num, dtype=torch.float, device=device)
+    yaw = torch.zeros(num, dtype=torch.float, device=device)
+    euler_angles = torch.stack([roll, pitch, yaw], dim=-1)
+    rotation_matrix = euler_angles_to_matrix(euler_angles, "XYZ")
+    quat = matrix_to_quaternion(rotation_matrix)
+    return quat
+
+@torch.jit.script
+def quat_diff_rad(a: torch.Tensor, b: torch.Tensor) -> torch.Tensor:
```
StoneT2000, line 332: no need for you to do it, just noting to myself i will move this code to a utility file in the future as it is pretty common.

PR #290, opened 2024-04-26
Tasks wth AI2 Thor assets

smtrived:


mani_skill/envs/tasks/assets/Shelving_Unit_307_1.glb
StoneT2000, line 1: is there no coacd.ply file for this?

smtrived: Just pushed it. Looks like it wasn't generating a .coacd.ply file because I had the shelf set to a kinematic object.

PR #366, opened 2024-06-05
Added RollBall env

guru-narayana:
A simple task where the objective is to push and roll a ball to a goal region at the other end of the table.
When testing with baseline PPO please use max_steps of 60, the ball takes time to roll


mani_skill/envs/tasks/tabletop/roll_ball.py
```diff
@@ -0,0 +1,166 @@
...
+    --------------
+    - the ball's xy position is randomized on top of a table in the region [0.2, 0.5] x [-0.4, 0.7]. It is placed flat on the table
+    - the target goal region is marked by a red/white circular target. The position of the target is randomized on top of a table in the region [-0.4, -0.7] x [0.2, -0.9]
+
+    Success Conditions
+    ------------------
+    - the ball's xy position is within goal_radius (default 0.1) of the target's xy position by euclidean distance.
+    """
+
+    SUPPORTED_REWARD_MODES = ["dense", "none"]
```
StoneT2000: this can be removed. all rewards are supported if u have success condition and dense reward

mani_skill/envs/tasks/tabletop/roll_ball.py
```diff
@@ -0,0 +1,166 @@
...
+        super().__init__(*args, robot_uids=robot_uids, **kwargs)
+
+    @property
+    def _default_sim_config(self):
+        return SimConfig(
+            gpu_memory_cfg=GPUMemoryConfig(
+                found_lost_pairs_capacity=2**25, max_rigid_patch_count=2**18
+            )
+        )
+
```
StoneT2000, line 57: please add a default sensor config. Since this is the same table top u can copy the pick cube.

StoneT2000 (changes requested):
Can you provide the exact command line script for ppo that I can just copy and paste?

Thank you.

guru-narayana:
Updated everything in accordance with your recommendations. 

Use the following command to run PPO

```
python examples/baselines/ppo/ppo.py --env_id="RollBall-v1"  --num_envs=1024 --update_epochs=8 --num_minibatches=32 --seed=100 --total_timesteps=100_00_000 --eval_freq=8 --num-steps=60 --num_eval_steps=60 --gamma 0.95
```

docs/source/tasks/table_top_gripper/index.md
```diff
@@ -282,3 +282,27 @@ One robot must push/give the cube on its side to the other side so the other rob
...
+**Randomizations:**
+- the ball's xy position is randomized on top of a table in the region [0.2, 0.5] x [-0.4, 0.7]. It is placed flat on the table
+- the target goal region is marked by a red/white circular target. The position of the target is randomized on top of a table in the region [-0.4, -0.7] x [0.2, -0.9]
+
+**Success Conditions:**
+-  the ball's xy position is within goal_radius (default 0.1) of the target's xy position by euclidean distance.
+:::
+
+<video preload="auto" controls="True" width="100%">
+<source src="https://github.com/haosulab/ManiSkill/raw/main/figures/environment_demos/RollBall-v1.mp4" type="video/mp4">
```
StoneT2000, line 338: I just checked the video. how did you generate it? It looks like the first frame is from another episode (asking about how you generated since this could be a bug with a tool of ours if you used our tools)

docs/source/tasks/table_top_gripper/index.md
```diff
@@ -282,3 +282,27 @@ One robot must push/give the cube on its side to the other side so the other rob
...
+![dense-reward][reward-badge]
+
+:::{dropdown} Task Card
+:icon: note
+:color: primary
+
+**Task Description:**
+A simple task where the objective is to push and roll a ball to a goal region at the other end of the table.
+
+**Supported Robots: (Panda, Panda)**
```
StoneT2000: this is not right, code says only the single agent "Panda" is supported. (Panda, Panda) indicates two pandas at once

docs/source/tasks/table_top_gripper/index.md
guru-narayana: I think I found where bug could be, when I am saving the trajectory from PPO with "--evaluate" its reset params is empty so that could be root cause. Is there a way to prevent this ?

StoneT2000: Ah so you are straight running the PPO code in the repo. I will check this

docs/source/tasks/table_top_gripper/index.md
guru-narayana: Fixed it

StoneT2000:
I am still training the environment @guru-narayana to check it solves in a reasonalbe time (the PPO script uses 100M steps which is a lot. If it solves in about an hour on a 3080 I will merge this in anyway). 

I recommend trying to tune the reward function a bit more (maybe a staged reward function can work better). Another point is that the reward gets the agent to be 0.05m behind the ball, but this can be suboptimal as you need to hit the ball at an angle.

StoneT2000:
Actually for the example ppo script did you mean to write 100_000_000, it says 100_00_000

guru-narayana:
> I am still training the environment @guru-narayana to check it solves in a reasonalbe time (the PPO script uses 100M steps which is a lot. If it solves in about an hour on a 3080 I will merge this in anyway).
> 
> I recommend trying to tune the reward function a bit more (maybe a staged reward function can work better). Another point is that the reward gets the agent to be 0.05m behind the ball, but this can be suboptimal as you need to hit the ball at an angle.

I made suggested modification and now the agent needs to be at 0.05m behind the ball and along the direction of the goal to get reward. I also made the reward function staged.

please use this modified command to test the environment.
`python ManiSkill/examples/baselines/ppo/ppo.py --env_id="RollBall-v1" --num_eval_envs=8 --num_envs=1024 --update_epochs=8 --num_minibatches=32  --total_timesteps=20_000_000 --eval_freq=10 --num-steps=80 --num_eval_steps=80  --gamma=0.95`

guru-narayana:
> Actually for the example ppo script did you mean to write 100_000_000, it says 100_00_000

100_00_000 was correct previously, but now please use the command in my recent comment for execution.

mani_skill/envs/tasks/tabletop/roll_ball.py
```diff
@@ -0,0 +1,179 @@
...
+            self.scene,
+            radius=self.goal_radius,
+            thickness=1e-5,
+            name="goal_region",
+            add_collision=False,
+            body_type="kinematic",
+        )
+
+    def _initialize_episode(self, env_idx: torch.Tensor, options: dict):
+        if self.reached_status.device != self.device:
```
StoneT2000: i don't understand why this needs to be done. How come reached status bool needs to be delayed one step for the reward function? Why not compute it in the reward function based on the tcp position to ball distance instantaneously?

StoneT2000 (changes requested):

StoneT2000:
Furthermore can you merge in the main branch? It seems you may have used a version on the main branch that had a small bug with the ManiSkillVectorEnv (apoligies for that). Things should run faster now / correctly. Otherwise i can verify this task works correctly, only that one small concern about the need to have a delayed boolean

mani_skill/envs/tasks/tabletop/roll_ball.py
guru-narayana: The "reached_status" keeps the track of envs that manged to reach behind the ball, post which the reward function returns the reward in the next stage, i.e rolling the ball by hitting.

I don't understand what you mean by delayed ? can you explain ?

I need to reset it every time the a env (could be partial envs too) gets a reset. so i kept it there. (also initializing in device is tricky because when i went through the source code the device implementation is still under progress so, I tried to do a future proof implementation)

line 165 to 172. of  mani_skill.envs.sapien_env
`        if num_envs > 1 or sim_backend == "gpu":
            if not physx.is_gpu_enabled():
                physx.enable_gpu()
            self.device = torch.device(
                "cuda"
            )  # TODO (stao): fix this for multi process support
        else:
            self.device = torch.device("cpu")
`

guru-narayana:
> Furthermore can you merge in the main branch? It seems you may have used a version on the main branch that had a small bug with the ManiSkillVectorEnv (apoligies for that). Things should run faster now / correctly. Otherwise i can verify this task works correctly, only that one small concern about the need to have a delayed boolean

Done

mani_skill/envs/tasks/tabletop/roll_ball.py
StoneT2000: you can safely use self.device. Each env object should be associated with one device.

for reached_status, I notice that you update this value after all the reward is done computing, hence it is delayed by one step. This does not make much sense (and it might slow RL down).

guru-narayana: Yeah that is a problem, i just committed a fix.

StoneT2000 (approved):

PR #375, opened 2024-06-09
[CSE 276F Submission] PokeCube-v1

xixinzhang:
A simple task where the objective is to poke a red cube with a peg and push it to a target goal position.

Solvable via ppo baseline:
```
python examples/baselines/ppo/ppo.py --env_id="PokeCube-v1" --num-eval-steps=50  --num_envs=1024 --update_epochs=8 --num_minibatches=32 --total_timesteps=5_000_000 --eval_freq=8 --num-steps=20
```


mani_skill/envs/tasks/tabletop/poke_cube.py
```diff
@@ -0,0 +1,192 @@
...
+        cube_q = self.cube.pose.q
+        cube_qmat = rotation_conversions.quaternion_to_matrix(cube_q)
+        cube_euler = rotation_conversions.matrix_to_euler_angles(cube_qmat, "XYZ")
+        angle_diff = torch.abs(peg_euler[:, 2] - cube_euler[:, 2])
+        is_peg_cube_aligned = angle_diff < 0.05
+
+        head_to_cube_dist = torch.linalg.norm(
+            self.peg_head_pos[..., :2] - self.cube.pose.p[..., :2], axis=1
+        )
+        is_peg_cube_touch = head_to_cube_dist <= self.cube_half_size + 0.005
```
StoneT2000: please rename is_peg_cube_touch to something like is_peg_cube_close. This condition does not check correctly if they are touching. Checking closeness is fine.

mani_skill/envs/tasks/tabletop/poke_cube.py
```diff
@@ -0,0 +1,192 @@
...
+            cube_xyz = torch.rand((b, 3)) * 0.2 - 0.1
+            cube_xyz[..., 0] = peg_xyz[..., 0] + self.peg_half_length + 0.1
+            cube_xyz[..., 2] = self.cube_half_size
+            cube_q = [1, 0, 0, 0]
+            cube_pose = Pose.create_from_pq(p=cube_xyz, q=cube_q)
+            self.cube.set_pose(cube_pose)
+            # initialize the goal region
+            goal_region_xyz = cube_xyz + torch.tensor([0.05 + self.goal_radius, 0, 0])
+            goal_region_xyz[..., 2] = 1e-3
+            goal_region_q = euler2quat(0, np.pi / 2, 0)
```
StoneT2000, line 114: can you make the task slightly more difficult by randomizing the quaternion for the cube? This seems too easy. I expect it to be still solvable with a bit of z-axis randomization

StoneT2000 (changes requested):

StoneT2000:
Thank you for your submission for the course. This task is nice and sufficiently interesting to be added. However could you make a different peg instead of the default provided magnet/blue red one? Since the peg does not need specific orientation

mani_skill/envs/tasks/tabletop/poke_cube.py
xixinzhang: I have modified the variable names containing "touch" incorrectly, thanks for your advice.

mani_skill/envs/tasks/tabletop/poke_cube.py
xixinzhang: Add z-axis randomization of the cube, rotate in the range [-pi/6, pi/6]

xixinzhang:
> Thank you for your submission for the course. This task is nice and sufficiently interesting to be added. However could you make a different peg instead of the default provided magnet/blue red one? Since the peg does not need specific orientation

The peg has been changed to single-color

mani_skill/envs/tasks/tabletop/poke_cube.py
```diff
@@ -0,0 +1,199 @@
...
+        head_to_cube_dist = torch.linalg.norm(
+            self.peg_head_pos[..., :2] - self.cube.pose.p[..., :2], axis=1
+        )
+        is_peg_cube_close = head_to_cube_dist <= self.cube_half_size + 0.005
+
+        is_peg_cube_fit = torch.logical_and(is_peg_cube_aligned, is_peg_cube_close)
+        is_peg_grasped = self.agent.is_grasping(self.peg)
+        close_to_table = torch.abs(self.peg.pose.p[:, 2] - self.peg_half_width) < 0.005
+        return {
+            "success": is_cube_placed & is_peg_cube_fit & close_to_table,
```
StoneT2000, line 160: There is still a bug here actually. The task sounds like you should move the cube to the target somehow, but why must the peg be close to the cube? close_to_table also does not make much sense. Eval videos are unintuitive as it looks like it is solved but this condition says it is not, better to just use the `is_cube_placed` condition only.

StoneT2000 (changes requested):

StoneT2000 (approved):
thank you, this looks good!

PR #376, opened 2024-06-09
[BugFix] allow set collision group bit to 0 in set_collision_group_bit

arth-shukla:


StoneT2000:
Wait why do we have two lines changing the bit, can this not be one line

StoneT2000:
Actually I think we should just also support bit being true or false. Easier to understand and read and this function is rarely called so we don't need extreme performance

arth-shukla:
I can change it to one line (i.e. `cg[group] = (cg[group & ~(1 << bit_idx)) | (bit << bit_idx)`), but it should be the same speed as the two lines above since either way we and then or

As for setting True/False, I can implement that by casting to int and doing the same operations.

Should I implement both of these changes?

StoneT2000:
Oh for some reason i thought collision groups were arrays, they are not. Yes just write as you did before and add support for booker with casting works fine

arth-shukla:
ok, pushed those changes. not sure if we should use different names than "bits" and such in the function name/args to abstract from the user (e.g. group/section/toggle instead of group/bit_idx/bit or smth), but that would require changes to each call of `set_collision_group_bit`

StoneT2000 (approved):

PR #377, opened 2024-06-09
[CSE 276F Submission] TranslatePeg-v1

xixinzhang:
The goal is to translate a peg to the target position on the table.

Solvable via ppo baseline:
```
# python examples/baselines/ppo/ppo.py --env_id="TranslatePeg-v1" --num-eval-steps=50  --num_envs=2048 --update_epochs=8 --num_minibatches=32 --total_timesteps=40_000_000 --eval_freq=8 --num-steps=20

```


StoneT2000:
Thank you for your submission for the course. Could you make the task a bit more difficult? It is a little bit too easy from first glance since the peg does not have randomized rotation on top of the table and it is always translates -0.2. I recommend randomization the rotation a bit (maybe +/- 30 degrees on the z-axis) and for translation it can translate x and y direction. Otherwise task is about as difficult as PushCube which is the easy/tutorial task.

StoneT2000:
Hi @xixinzhang this task is a bit too simple and similar in terms of dynamics to some other tasks and is a bit too easy. I will close the PR for now in favor of your other PR which is more likely to be merged. Feel free to discuss on Piazza for more clarifications.

PR #405, opened 2024-07-04
[Feature] add G1 robot

matheecs:


StoneT2000:
Thank you for your contribution in adding the new G1 robot! I will review this as soon as I can. Most likely though we might want to move the robot assets to under our organization and document any changes made to the original URDF but the rest should be fine.

mani_skill/utils/download_asset.py
```diff
@@ -186,6 +186,11 @@ def initialize_extra_sources():
         url="https://github.com/haosulab/ManiSkill-UnitreeH1/archive/refs/tags/v0.1.0.zip",
         target_path="robots/unitree_h1",
     )
+    DATA_SOURCES["unitree_g1"] = DataSource(
+        source_type="robot",
+        url="https://github.com/matheecs/ManiSkill-UnitreeG1/archive/refs/tags/v0.1.0.zip",
```
StoneT2000: I have forked this to our own repository. Can you change the link to `https://github.com/haosulab/ManiSkill-UnitreeG1/archive/refs/tags/v0.1.0.zip`

StoneT2000 (changes requested):
All the code looks mostly good. There are a few requests before I can finally merge this in. 

For one, the G1 URDF is good but it is too complicated and not suitable for RL work (it runs too slowly). Is there a simplified G1 robot URDF? You can check out the H1 robot URDF we have here: https://github.com/haosulab/ManiSkill-UnitreeH1/tree/main/urdf

A simplified URDF is necessary to make simulation run much faster (fewer collision shapes and faster solving). 

It is okay if a simplified G1 URDF is not available. If it is not available, then the only other change requested here is to remove the G1 stand task you added and the example ppo script added to the readme.

mani_skill/utils/download_asset.py
matheecs: Sure. Please pull the latest update from the original repository (with the new simplified model added [g1_simplified.urdf](https://github.com/matheecs/ManiSkill-UnitreeG1/blob/main/g1_simplified.urdf)).

matheecs:
I created a simplified URDF, [g1_simplified.urdf](https://github.com/matheecs/ManiSkill-UnitreeG1/blob/main/g1_simplified.urdf), to improve simulation speed. The training demo looks like this:

https://github.com/user-attachments/assets/e32e2495-56f4-4d6b-9eae-af0466a38585

Note: Please pull the latest update from the [original repository](https://github.com/matheecs/ManiSkill-UnitreeG1).

StoneT2000:
Maybe a bit too simplified. Certainly runs fine but since the simplification removes everything but meshes for the feet / legs, can you rename unitree_g1_simplified to unitree_g1_simplified_legs? Same for the URDF file name to make it clear. I think some users may find it confusing why the simplified model can fall through the floor a bit.

After that then happy to merge.

matheecs:
I renamed unitree_g1_simplified to unitree_g1_simplified_legs and updated the URDF name. Thanks.

Note: Please pull the latest update from the [original repository](https://github.com/matheecs/ManiSkill-UnitreeG1).

StoneT2000:
lgtm, thanks for your contribution!

StoneT2000 (approved):

PR #411, opened 2024-07-10
[Feature] Updated with voxel observation mode. Added observation mode config supports

astonastonaston:
* Added the voxel observation mode allowing the convertion from RGBD sensor data to voxels. 
* Updated docs with some explanations, examples, and visualized images
* Supported observation mode config so that observation hyper-parameters can be passed in during environment initializations


mani_skill/utils/visualization/voxel_visualizer.py
```diff
@@ -0,0 +1,201 @@
+# Borrowed and modified from https://github.com/stepjam/ARM/blob/main/arm/utils.py
+
+import numpy as np
+import pyrender
+import torch
+import trimesh
+from pyrender.trackball import Trackball
+# from rlbench.backend.const import DEPTH_SCALE
```
StoneT2000: remove import from rlbench

mani_skill/examples/demo_vis_voxel.py
StoneT2000, line 1: nice that you added this

docs/source/user_guide/concepts/observation.md
```diff
@@ -92,6 +92,71 @@ For a quick demo to visualize pointclouds, you can run
 python -m mani_skill.examples.demo_vis_pcd -e "PushCube-v1"
 ```
 
+### voxel
+This observation mode has the same data format as the [sensor_data mode](#sensor_data), but all sensor data from cameras are removed and instead a new key is added called `voxel_grid`.
+
+To use this observation mode, a dictionary of observation config parameters is required to be passed in via obs_mode_config during environment initializations (gym.make()). It should contain the following voxelization config hyperparameters:
```
StoneT2000: can you include documentation at the top of the document discussing about how some observation modes permit observation configurations (to override any defaults)? And how to do so?

For some tasks we might have some default voxel grid configs (to support easy benchmarking of methods like peract by simply doing gym.make(env_id) instead of having to have the user specify some set of voxel configs).

StoneT2000 (changes requested):
Great work so far and looks very thorough. Just a few comments

astonastonaston:
Now default voxelization configs are added. Voxelization can be done by just gym.make("PushCube-v1", obs_mode="voxel").

mani_skill/envs/tasks/tabletop/place_sphere.py
```diff
@@ -62,25 +62,23 @@ def _default_sim_config(self):
 
     @property
     def _default_sensor_configs(self):
-        pose = sapien_utils.look_at(eye=[0.3, 0, 0.2], target=[-0.1, 0, 0])
```
StoneT2000, line 72: why is the original camera pose changed?

mani_skill/envs/tasks/quadruped/quadruped_reach.py
```diff
@@ -63,6 +63,13 @@ def _default_human_render_camera_configs(self):
                 # mount=self.agent.robot.links[0],
             )
         ]
+    
+    @property
```
StoneT2000: remove this. Typically voxels aren't used for this task anyway and also we don't know if this setting makes sense

StoneT2000 (changes requested):
Lets limit the default voxel configs to just the tasks in the tabletop folder for now so remove the other defaults. We don't know what voxel settings are reasonable for the other tasks (and users won't use them probably). No default is better than possible errorneous default

mani_skill/envs/tasks/tabletop/place_sphere.py
astonastonaston: It's just to align with other tabletop tasks' camera poses like PushCube/PickCube/StackCube's default poses (they use the same poses)

astonastonaston:
Now the voxel obs config for non-tabletop tasks are removed. The demos' segmentation supports are modified a bit for better visualizations.

Issue #419, opened 2024-07-12
[Question] Color randomization as in ManiSkillv2


alexcbb:
Hello,

I'm currently trying to create an environment with different objects and random colors at each episode. I've searched on how to do that and saw an old documentation of ManiSkillv2 but it seems that it does not work this way now : 
https://github.com/haosulab/ManiSkill/blob/v0.5.3/examples/tutorials/customize_environments.ipynb (in Domain Randomization part)

Would you have a proper way to do it now ? I've been searching a way to access the visual bodies of the actor but didn't find anything.

Thank in advance !

StoneT2000:
For object randomization you can check how PickSingleYCB task works. Note that in GPU simulation by default reconfiguration only happens once as it is a bit slow.

For color and texture randomization there is a way to do it, i'll provide some examples like before in some documentation later.

alexcbb:
> For object randomization you can check how PickSingleYCB task works. Note that in GPU simulation by default reconfiguration only happens once as it is a bit slow.
> 
> For color and texture randomization there is a way to do it, i'll provide some examples like before in some documentation later.

Yes, I've already have the code for object randomization, my issue was more about the color randomization, I've got a solution currently that is only working on CPU and require deleting the actors. But if a better solution exists that would be perfect.

StoneT2000:
For color/texture randomization of already built objects you can check out the code for PushT: https://github.com/haosulab/ManiSkill/blob/main/mani_skill/envs/tasks/tabletop/push_t.py#L39-L46

Essentially during environment building/reconfiguration, you can modify at runtime the color of objects. For example if you are building a cube object, you use

```
builder.add_box_visual(
        half_size=[half_size] * 3,
        material=sapien.render.RenderMaterial(
            base_color=color,
        ),
    )
builder.set_scene_idxs([scene_idx]) # optional if you want to only build it in the scene_idx parallel environment
```
to add a box visual. You can pick a color here during reconfiguration (modify the _load_scene function of your environment). If you want a different color/texture per parallel environment, then make sure that when using the builder, only build the object in one of the sub-scenes, and build it again with a different texture in other sub-scenes

Note that you must reconfigure your environment or else these color changes will not work / there will be errors that show up. You cannot change textures (easily) after reconfiguration.

[...]

PR #420, opened 2024-07-12
Fix issue when trying to delete actor from scene

alexcbb:
When trying to call the `remove_actor` function on a created actor, I obtained the following error : 
```
TypeError: remove_entity(): incompatible function arguments. The following argument types are supported:
    1. (self: sapien.pysapien.Scene, entity: sapien.pysapien.Entity) -> None

Invoked with: <sapien.wrapper.scene.Scene object at 0x777ea016f4d0>, <distractor_0: struct of type <class 'mani_skill.utils.structs.actor.Actor'>; managing 1 <class 'sapien.pysapien.Entity'> objects>
```
I adapted the code in order to get all entities related to the actor in order to delete them, as it was already done for the articulations.

### DIFF
--- a/mani_skill/envs/scene.py
+++ b/mani_skill/envs/scene.py
@@ -129,13 +129,13 @@ class ManiSkillScene:
     ):
         return physx.PhysxMaterial(static_friction, dynamic_friction, restitution)
 
-    def remove_actor(self, actor : Actor):
+    def remove_actor(self, actor: Actor):
         if physx.is_gpu_enabled():
             raise NotImplementedError(
                 "Cannot remove actors after creating them in GPU sim at the moment"
             )
         else:
-            entities = [l for l in actor_objs]
+            entities = [l for l in actor._objs]
             for e in entities:
                 self.sub_scenes[0].remove_entity(e)
 
--- a/mani_skill/envs/scene.py
+++ b/mani_skill/envs/scene.py
@@ -135,9 +135,7 @@ class ManiSkillScene:
                 "Cannot remove actors after creating them in GPU sim at the moment"
             )
         else:
-            entities = [l for l in actor._objs]
-            for e in entities:
-                self.sub_scenes[0].remove_entity(e)
+            self.sub_scenes[0].remove_entity(e for e in actor._objs)
 
     def remove_articulation(self, articulation: Articulation):
         if physx.is_gpu_enabled():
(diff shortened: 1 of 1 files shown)


StoneT2000:
Thank you for your contribution. Just want to point out that you cannot actually delete an actor from the scene in the GPU sim so be careful about using this function. Curious, what is the intended goal of removing an actor in your use case?

mani_skill/envs/scene.py
```diff
@@ -129,13 +129,15 @@ def create_physical_material(
     ):
         return physx.PhysxMaterial(static_friction, dynamic_friction, restitution)
 
-    def remove_actor(self, actor):
+    def remove_actor(self, actor : Actor):
```
StoneT2000: python format is wrong here. Please follow the CONTRIBUTING.md instructions for setting up pre-commit which comes with black and isort for formatting.

mani_skill/envs/scene.py
```diff
@@ -129,13 +129,15 @@ def create_physical_material(
...
         if physx.is_gpu_enabled():
             raise NotImplementedError(
                 "Cannot remove actors after creating them in GPU sim at the moment"
             )
         else:
-            self.sub_scenes[0].remove_entity(actor)
+            entities = [l for l in actor_objs]
```
StoneT2000: this code does not work. actor_objs is undefined.

StoneT2000 (changes requested):

alexcbb:
> Thank you for your contribution. Just want to point out that you cannot actually delete an actor from the scene in the GPU sim so be careful about using this function. Curious, what is the intended goal of removing an actor in your use case?

Hello, the use case is in fact related to the following issue : https://github.com/haosulab/ManiSkill/issues/419#issuecomment-2226677176 ; I wanted to find a way to randomize the color of objects at the reset of an episode but could'nt found a way to do it properly. I then found a way to remove the objects and just create them as new random shapes.

alexcbb:
@StoneT2000 
Excuse me for my bad formatting, this is my first contribution on an open-source project, I will try to be more vigilent next time.
I've indeed forgot a point in the code, it should be fixed now with the new commit (for the formatting and the wrong var name)

mani_skill/envs/scene.py
```diff
@@ -129,13 +129,15 @@ def create_physical_material(
...
 
-    def remove_actor(self, actor):
+    def remove_actor(self, actor: Actor):
         if physx.is_gpu_enabled():
             raise NotImplementedError(
                 "Cannot remove actors after creating them in GPU sim at the moment"
             )
         else:
-            self.sub_scenes[0].remove_entity(actor)
+            entities = [l for l in actor._objs]
```
StoneT2000: small nit, this can be improved by doing [self.sub_scenes[0].remove_entity(x) for x in actor._objs]

alexcbb: @StoneT2000 The last commit fixed this, I also applied it for the following function that performed the same operation

StoneT2000 (changes requested):
Double check the code runs. There is a bug in this code. remove_entity does not support a generator (or list) as an argument.

PR #431, opened 2024-07-16
Initialize _target_qvel as self.qvel when resetting if it is None

sean1295:
Currently control_mode="pd_joint_pos_vel" doesn't work.

This should fix that.


mani_skill/agents/controllers/pd_joint_pos_vel.py
```diff
@@ -23,7 +23,10 @@ def _initialize_action_space(self):
 
     def reset(self):
         super().reset()
-        self._target_qvel[self.scene._reset_mask] = torch.zeros_like(
+        if self._target_qvel is None:
+            self._target_qvel = self.qvel.clone()
+        else:
+            self._target_qvel[self.scene._reset_mask] = torch.zeros_like(
             self._target_qpos[self.scene._reset_mask], device=self.device
```
StoneT2000: small nit, please run the pre-commit formatters (black and isort). The formatting here looks incorrect

StoneT2000 (changes requested):

PR #437, opened 2024-07-17
[Feature] Add RLPD Baseline

StoneT2000:
RLPD is sota sample-efficient online imitation learning method (without using state reset).


StoneT2000:
note that the recorded metrics / config is not aligned with other baselines yet

PR #440, opened 2024-07-18
Control Hopper Env, Stand + Hop

Xander-Hinrichsen:
1. Added hopper xml, hopper environment, hopper stand and hop tasks
MS-HopperStand-v1 & MS-HopperHop-v1

2. Made small changes to the following files
mani_skill/envs/sapien_env.py - added ability to add articulation mount for render only camera - this allows the ppo evaluation videos to be mounted to robot articulation (necessary to view hopper throughout episode and it's movement) - doesn't make any difference for _human_render_cameras if entity_uid in camera config isn't specified - as usual for all previous environments
mani_skill/utils/building/ground.py - added additional kwargs to ground builder to allow rectangular grounds (no longer just square) and ground xy offsets (both necessary for dm_control planar agent ground visual in mujoco planar agent environments style) - no difference for this api if these additional kwargs aren't specified - as usual for all previous environments
index - task documentation

3. Made large changes to the following files
mani_skill/envs/utils/rewards/common.py - value at margin wasn't implemented, and the sigmoidal reward fns weren't guaranteed to return rewards between 0 and 1! - made changes s.t. it's equivalent to dm_control reward tolerance

These changes make it much easier for implementation of other planar agents (like walker and cheetah), but in my experience, there are issues (like known and documented bugs in mjcf_loader.py and joint tuning difficulty) that make these other environments harder to implement

image

image

environment requires further tuning to more resemble dm_control hopper

https://github.com/user-attachments/assets/3412df4c-cb86-401a-ba19-e0209167797b

https://github.com/user-attachments/assets/a791bd42-e712-4ba3-b1e6-61cefa637daa

args to recreate: 
Hop 
for i in {1..3}; do python ppo.py --exp_name="hopseed${i}" --env_id="MS-HopperHop-v1" --num_envs=2048 --update_epochs=8 --num_minibatches=32   --total_timesteps=20_000_000 --eval_freq=10 --num_eval_steps=100 --num_steps=100 --gamma=0.999 --seed=${i}; done
Stand
for i in {1..3}; do python ppo.py --exp_name="standseed${i}" --env_id="MS-HopperStand-v1" --num_envs=2048 --update_epochs=8 --num_minibatches=32   --total_timesteps=20_000_000 --eval_freq=10 --num_eval_steps=100 --num_steps=100 --gamma=0.99 --seed=${i}; done
HopRGB singleseed 
python ppo_rgb.py --exp_name="rgbhopseed1" --env_id="MS-HopperHop-v1" --num_envs=1024 --update_epochs=8 --num_minibatches=32   --total_timesteps=20_000_000 --eval_freq=10 --num_eval_steps=100 --num_steps=100 --gamma=0.99 --seed=1


mani_skill/envs/tasks/control/hopper.py
```diff
@@ -0,0 +1,266 @@
...
+    def touch(self, link_name):
+        """Returns function of sensor force values"""
+        force_vec = self.agent.robot.get_net_contact_forces([link_name])
+        force_mag = torch.linalg.norm(force_vec, dim=-1)
+        return torch.log1p(force_mag)
+
+    # dm_control also includes foot pressures as state obs space
+    def _get_obs_extra(self, info: Dict):
+        obs = dict(
+            toe_touch=self.touch("foot_toe"),
```
StoneT2000, line 208: does dm control with visual data input ever use state information? This data will be part of the observation (along with _get_obs_agent())

mani_skill/envs/tasks/control/hopper.py
```diff
@@ -0,0 +1,266 @@
...
+        super().__init__(*args, robot_uids=robot_uids, **kwargs)
+
+    @property
+    def _default_sim_config(self):
+        return SimConfig(
+            gpu_memory_cfg=GPUMemoryConfig(
+                found_lost_pairs_capacity=2**25, max_rigid_patch_count=2**18
+            ),
+            scene_cfg=SceneConfig(
+                solver_position_iterations=15, solver_velocity_iterations=15
```
StoneT2000: why these configs for solver? There is no need to try and match mujoco solver configurations since mujoco is a different physics engine compared to what we use which is physx

mani_skill/envs/tasks/control/hopper.py
```diff
@@ -0,0 +1,266 @@
...
+            ),
+            # sim_freq=50,
+            control_freq=25,
```
StoneT2000, line 118: how is control_freq here decided, was it to try and match dm-control?

mani_skill/envs/sapien_env.py
```diff
@@ -620,9 +620,14 @@ def _setup_sensors(self, options: dict):
...
         self._human_render_cameras = dict()
         for uid, camera_cfg in self._human_render_camera_configs.items():
+            if uid in self._agent_sensor_configs:
+                articulation = self.agent.robot
+            else:
+                articulation = None
             self._human_render_cameras[uid] = Camera(
                 camera_cfg,
                 self.scene,
+                articulation=articulation,
```
StoneT2000: actually we should be deprecating the use of a link string id to specify where to mount as it requires the implicit assumption that we fill in the correct articulation to look for the corresponding link string id.

Better code is to directly reference the link object in the camera config (see later comments). Hence these new lines should be removed.

mani_skill/envs/tasks/control/hopper.py
```diff
@@ -0,0 +1,266 @@
...
+            # replicated from xml file
+            CameraConfig(
+                uid="cam0",
+                pose=Pose.create_from_pq([0, -2.8, 0], euler2quat(0, 0, np.pi / 2)),
+                width=128,
+                height=128,
+                fov=70 * (np.pi / 180),
+                near=0.01,
+                far=100,
+                entity_uid="torso_dummy_1",
```
StoneT2000: use `mount=self.robot.links_map["torso_dummy_1"]` instead. This is explicit and better as it references the actual link to mount on. Remove entity_uid (we should be deprecating the use of that in the future).

mani_skill/envs/tasks/control/hopper.py
```diff
@@ -0,0 +1,266 @@
...
+            scene_cfg=SceneConfig(
+                solver_position_iterations=15, solver_velocity_iterations=15
+            ),
+            # sim_freq=50,
+            control_freq=25,
+        )
+
+    @property
+    def _default_sensor_configs(self):
+        return self.agent._sensor_configs
```
StoneT2000: this shouldn't be necessary, the agent sensors are already part of the environment's sensors.

mani_skill/envs/tasks/control/hopper.py
```diff
@@ -0,0 +1,266 @@
...
+
+    @property
+    def _default_human_render_camera_configs(self):
+        return self.agent._sensor_configs
```
StoneT2000: It would be better to just copy the camera config for the agent sensor and put it here. You also should change the camera name and make the default height/width of the camera to be larger (512x512)

mani_skill/envs/tasks/control/hopper.py
StoneT2000: Actually, I would request the camera being defined here instead of part of the agent. This is because the camera is a third-party camera not really part of the robot (although it is focused on the robot). This is mostly just maniskill convention. There are tasks where perhaps the camera should not be centered on the agent, in which case then when using the hopper agent one has to write code to get rid of that camera.

mani_skill/envs/tasks/control/hopper.py
```diff
@@ -0,0 +1,266 @@
...
+
+    @property
+    def _sensor_configs(self):
+        return [
+            # replicated from xml file
+            CameraConfig(
+                uid="cam0",
+                pose=Pose.create_from_pq([0, -2.8, 0], euler2quat(0, 0, np.pi / 2)),
+                width=128,
+                height=128,
```
StoneT2000: is 128x128 correct? I recall dm-control uses smaller images

mani_skill/envs/tasks/control/hopper.py
```diff
@@ -0,0 +1,266 @@
...
+    def _default_human_render_camera_configs(self):
+        return self.agent._sensor_configs
+
+    def _load_scene(self, options: dict):
+        loader = self.scene.create_mjcf_loader()
+        articulation_builders, actor_builders, sensor_configs = loader.parse(MJCF_FILE)
+        for a in actor_builders:
+            a.build(a.name)
+
+        # load in the ground
```
StoneT2000: for hopper (and also cartpole). I recommend we add a massive long white wall behind so that the rendered videos look a bit nicer with color 0.3, 0.3, 0.3 (this will match what ray-traced videos will look like since they include the ambient lighting  which is 0.3 0.3 0.3). Moreover with a big background wall if anyone wants to use depth image data they can and it won't be full of 0s (indicating no depth info since it goes to infinity).

StoneT2000 (changes requested):

mani_skill/envs/tasks/control/hopper.py
Xander-Hinrichsen: I am now explicitly overriding _get_obs_state_dict to include entire state (touch + proprioception), so the touch information isn't allowed for rgbd obs mode

mani_skill/envs/tasks/control/hopper.py
Xander-Hinrichsen: changed to solver_position_iteration from default 15 to 4, for sim efficiency (15 was unnecessary), and vel iterations stays default value of 1

mani_skill/envs/tasks/control/hopper.py
Xander-Hinrichsen: control_freq was set so the ratio of sim_freq to control is 4:1 (same as dm_control hopper), code now explicitly has sim_freq=100

mani_skill/envs/sapien_env.py
Xander-Hinrichsen: fixed

mani_skill/envs/tasks/control/hopper.py
Xander-Hinrichsen: fixed :+1:

mani_skill/envs/tasks/control/hopper.py
Xander-Hinrichsen: fixed :+1: :+1:

mani_skill/envs/tasks/control/hopper.py
Xander-Hinrichsen: fixed :+1: :+1: :+1:

mani_skill/envs/tasks/control/hopper.py
Xander-Hinrichsen: i found that images are 240 height, 320 width:

```
import dm_control.suite as suite
from dm_control.suite.wrappers import pixels
import numpy as np

env = suite.load("hopper", "hop")
env_and_pixels = pixels.Wrapper(env, pixels_only=True)

_, _, _, obs = env_and_pixels.reset()

print(obs['pixels'].shape) # prints (240, 320, 3)
```
this seems large, should we use the same dims?

mani_skill/envs/tasks/control/hopper.py
Xander-Hinrichsen: created new scene builder in mani_skill/utils/scene_builder/control/planar, for re-use for the other planar agents
new eval visual (512x512): 

https://github.com/user-attachments/assets/4a7cc102-52e0-4660-9356-cd0927878f33

mani_skill/envs/tasks/control/hopper.py
Xander-Hinrichsen: dm_control uses pixel wrapper, which allows for user to choose either entire state + pixels or only pixels

mani_skill/envs/tasks/control/hopper.py
Xander-Hinrichsen: added wall to cartpole.py file directly, since it doesn't fit with planar scene builder or warrant its own scene builder

https://github.com/user-attachments/assets/8928292d-c58d-4b56-857c-126bdcf0ee5a

Xander-Hinrichsen:
modified xml file hip range because original hopper robot is bugged - its hip range allows self intersection, getting the robot stuck in a certain position, with small probability on initialization:

dm_control_hopper_bug

mani_skill/envs/tasks/control/hopper.py
StoneT2000: ok we might want to play around with lighting in the future to make these look a bit brighter, but no need for now.

mani_skill/envs/tasks/control/hopper.py
StoneT2000: Actually turns out people seem to use different resolutions. Let's just stick with 128x128 for now. It runs fast and is sufficient.

Just checked TDMPC2 code and they use 384x384. So not sure how this is picked usually.

StoneT2000:
Is the ppo_rgb.py code updated? Seem to run into a few errors due to evaluating 100 steps instead of the max 600.

mani_skill/envs/tasks/control/hopper.py
StoneT2000: Ok looking at what seems to be standard practice, DrQ and TDMPC both do not use state at all in training, e.g. https://github.com/nicklashansen/tdmpc2/blob/5f6fadec0fec78304b4b53e8171d348b58cac486/tdmpc2/envs/wrappers/pixels.py#L8

Given the tasks are not that hard, we can do image only ppo as well. The way I will do it is to fix the flatten rgbd obsevation wrapper to be like the pixel wrapper, and you just choose if you want rgb, depth, and / or state, and it will handle it accordingly.

StoneT2000: re this, merge the main branch into your PR. I have updated code to do this exactly. The new ppo rgb script is this:

```
python ppo_rgb.py --exp_name="rgbhopseed1" --env_id="MS-HopperHop-v1" --num_envs=512 --update_epochs=8 --num_minibatches=32 --total_timesteps=20_000_000 --eval_freq=10 --num_eval_steps=600 --num_steps=50 --gamma=0.99 --seed=1 --no_include_state
```
(the hyperparams might not be well tuned).

Xander-Hinrichsen:
> Is the ppo_rgb.py code updated? Seem to run into a few errors due to evaluating 100 steps instead of the max 600.

yes, I added ppo_rgb 1 line change to get it to work in new commit - there was no check that success exists in final_info dict in ppo_rgb.py when tensorboard writing, while there is a check in ppo.py

for i in {1..3}; do python ppo_rgb.py --exp_name="pixels_hopseed${i}" --env_id="MS-HopperHop-v1"   --num_envs=256 --update_epochs=8 --num_minibatches=8   --total_timesteps=25_000_000 --num-steps=100 --num_eval_steps=600 --gamma=0.995 --seed=${i}; done

image


performance is worse without the pressure sensor information

StoneT2000:
It is hard to tell if it's "worse" since this is a different return scale now? Also i tried running again but the performance after 5M steps looks like this:

https://github.com/user-attachments/assets/a8a1cb45-c699-41f9-a02d-27e225c1f893


Looks like it learns to flip forward to go fast now, instead of the original behavior.

StoneT2000:
My script:

python ppo_rgb.py --exp_name="rgbhopseed1" --env_id="MS-HopperHop-v1" --num_envs=512 --update_epochs=8 --num_minibatches=32 --total_timesteps=20_000_000 --eval_freq=15 --num_eval_steps=600 --num_steps=100 --gamma=0.99 --seed=1 --no-include-state

(merge main for this to work)
This seems to perform better than the script you ran with these hyperparams. It does show some sign that it might learn the "right" behavior (it could also be that our hopper setup affords more optimal flipping than mujocos) 

image

Maybe behavior will be removed if the agent does not have so much power behind its actions. I can try and experiment as well.

Xander-Hinrichsen:
This has been my experience with the damping, stiffness, and delta being very important for desired behvior. If the damping especially is too low, then the hopper is unable to upright itself, if it's too high, then it can instead flip to its feet too easily.

StoneT2000:
can you somehow revert to the settings that got the first video that was working? or was it only working from state inputs

Xander-Hinrichsen:
it was only working from state inputs, these are the same controller settings that produced the first videos

StoneT2000:
Arguably though with the flipping it seems like it achieves the same distance if not further than the standard "behavior". I would then say this is more of a result of the reward function. Since its not possible to get the same exact simulation as mujoco, maybe we can just penalize flipping (just penalize the x/y axis rotation velocity should be sufficient, whichever is the axis pointing to the camera). It's fine if our reward function is slightly different to mujocos.

I find many other repos tackling locomotion like isaac lab / orbit will often add reward terms not meant to make the robot run faster, but to have a smoother gait, which is what we will be doing here essentially.

Xander-Hinrichsen:
okay, that sounds good, I will try some experiments with discouraging flipping reward

Xander-Hinrichsen:
> Maybe behavior will be removed if the agent does not have so much power behind its actions. I can try and experiment as well. 

No need to try to experiment, I have found that y axis rotation bounds for reward works well, will run tests and report here later tonight

edit: update - It worked well for a quick check on state obs - to ensure no flipping with large deltas. However, I've tried many, many different things, and unfortunately, any additional reward doesn't fix the problem - in fact, I believe the flipping was originally a local min solution, where hopping would result in greater reward, since the flipping doesn't always activate the standing reward portion of the hopping reward

currently testing out non delta joint pos

StoneT2000:
> > Maybe behavior will be removed if the agent does not have so much power behind its actions. I can try and experiment as well.
> 
> No need to try to experiment, I have found that y axis rotation bounds for reward works well, will run tests and report here later tonight
> 
> edit: update - It worked well for a quick check on state obs - to ensure no flipping with large deltas. However, I've tried many, many different things, and unfortunately, any additional reward doesn't fix the problem - in fact, I believe the flipping was originally a local min solution, where hopping would result in greater reward, since the flipping doesn't always activate the standing reward portion of the hopping reward
> 
> currently testing out non delta joint pos

what do you mean non delta joint pos? all of them should use that controller

Xander-Hinrichsen:
> what do you mean non delta joint pos? all of them should use that controller
I tried pd_joint_pos instead of pd_joint_delta_pos is what I meant, didn't work though

I've diagnosed the problem:
Originally, I tried to implement the robot center of mass x velocity, but it didn't work, so I just used the xvelocity of the slider joint holding the hopper in the xz plane (see in current implementation of subtreelinvelx function).

Since the entire robot's center of mass x velocity isn't taken into consideration (and only the torso xvel which is attached to the slider), the robot isn't penalized for flipping in a large circle as long as its torso continues to move in the +x world frame direction

Here is original robot center of mass velocity function, which is slow and doesn't work
```
    @property  # dm_control mjc function adapted for maniskill
    def subtreelinvelx(self):
        """Returns linvel x component of robot"""
        total_mass = 0.0
        vels = []
        links = self.agent.robot.get_links()
        # skip first three dummy links
        for i in range(3,len(links)):
            vel = links[i].get_linear_velocity() # vel is (b,3)
            mass = links[i].get_mass()[0]
            vels.append(vel * mass.item())
            total_mass += mass.item()
        vels = torch.stack(vels, dim=0) # (#links, b, 3)
        com_vel = vels.sum(dim=0) / total_mass # (b, 3)
        return com_vel[:,0]
```
it seems like the get_linear_velocity() function doesn't return the center of mass linear velocity of the links?, which is causing the issue.

If calculating the center of mass velocity for an entire robot is non-trivial, this is an issue for hopper + all planar + all humanoid dm_control task implementations, since they com velocity as reward

StoneT2000:
Lemme get back to you. I think I also just found a bug with the GPU sim linear and angular velocities. CPU sim is correct but on GPU sim they might flipped. Working with @fbxiang now to debug it.

StoneT2000:
Actually you can use the torso link. It should work. That is the link they are reading the center of mass velocities from. All velocity data is center of mass actually. 

e.g.

```
link = self.agent.robot.links_map["torso"]
print(link.linear_velocity, link.angular_velocity)
```

Also merge in main before doing more experiments. I just fixed #450 merged in #452 , so the link velocities used or com velocities are wrong lol

Thinking also more, maybe it is also better to just penalize the y-axis rotation, instead of velocity (its okay to go fast, just don't flip over).

mani_skill/utils/scene_builder/control/planar/scene_builder.py
```diff
@@ -0,0 +1,56 @@
...
+            ),
+        )
+        self.wall.build_static(name="wall")
+        self.scene_objects: List[sapien.Entity] = [self.ground, self.wall]
+
+    def initialize(self, env_idx: torch.Tensor):
+        b = len(env_idx)
+        robot = self.env.agent.uid
+        if robot == "hopper" or robot == "walker":
+            # qpos sampled same as dm_control, but ensure no self intersection explicitly here
```
StoneT2000: recommend moving robot specific initializations to the task itself. This way users can more easily customize and modify randomizations in just that one task file

StoneT2000 (changes requested):
small change requested i didnt catch earlier. See other comments above

Xander-Hinrichsen:
**The results have large variance because in both hop and stand tasks, in seed 2, ppo learns to straighten its legs out too early, and doesn't explore tucking them in when hopper is on its belly, in order to get up - ppo then gets stuck at this local optimum in seed 2 - This is a a result of lack of exploration in ppo, not a fault of the environment, as shown below

### Standing results and command: 

https://github.com/user-attachments/assets/63e0408f-e064-4a84-9c89-36e65690f2de

image

for i in {1..3}; do python ppo.py --exp_name="standseed${i}" --env_id="MS-HopperStand-v1" --num_envs=4096 --update_epochs=8 --num_minibatches=32 --total_timesteps=30_000_000 --eval_freq=10 --num_eval_steps=600 --num_steps=100 --gamma=0.99 --ent_coef=0.01 --seed=${i}; done

### Hopping results and command:

https://github.com/user-attachments/assets/e6675fdc-082c-434d-94fc-567a2c35b24e

image

for i in {1..3}; do python ppo.py --exp_name="hopseed${i}" --env_id="MS-HopperHop-v1" --num_envs=4096 --update_epochs=8 --num_minibatches=32 --total_timesteps=60_000_000 --eval_freq=10 --num_eval_steps=600 --num_steps=100 --gamma=0.99 --ent_coef=0.01 --seed=${i}; done

### RGB Hop results and command:


https://github.com/user-attachments/assets/8e6945d0-011d-4432-aa9b-d60c6404c09b


image

python ppo_rgb.py --exp_name="pixelhop" --env_id="MS-HopperHop-v1" --num_envs=512 --update_epochs=8 --num_minibatches=32 --total_timesteps=40_000_000 --eval_freq=15 --num_eval_steps=600 --num_steps=100 --gamma=0.99 --seed=1 --no-include-state

### Local Optima in PPO training

seed 2 ppo local optimum:
ppo learns to straighten its legs out no matter what, and doesn't overcome this local optimum during training in seed 2

https://github.com/user-attachments/assets/a51c9a28-a76f-4b1a-aa51-92f9f3f486dc

seed 1 and 3 don't have this issue: 

seed 1 stand:

https://github.com/user-attachments/assets/56d488ff-e920-4226-bee6-480a3192d90d


seed 3 stand:

https://github.com/user-attachments/assets/63e0408f-e064-4a84-9c89-36e65690f2de

**and even if the hopper learns to flip, it is a local minimum as it will eventually learn that hopping results in more rewards than flipping, this is because the center of mass of entire robot body reward correctly rewards hopping more than flipping

seed 3 of hopping command above first learns to flip, then later learns better how to hop to maximize the reward, eventually forgetting the flipping behavior:** 
remnant of flipping behavior shown during middle of training:


https://github.com/user-attachments/assets/a9675146-d10a-4c5c-9274-77cf24fde986

StoneT2000 (approved):

PR #452, opened 2024-07-25
[BugFix] Fix bug where linear/angular velocities are swapped in GPU sim vs CPU sim

StoneT2000:
After some debugging @fbxiang and I found out (astonishingly) that physx exposed GPU buffer data of link velocities are in a different order to their CPU sim counterpart...


StoneT2000:
closes #450

Issue #526, opened 2024-08-26
Dependency conflict in tdmpc2 baseline with maniskill dep update 


chenyenru:
To reproduce:

``` bash
conda env create -f examples/baselines/tdmpc2/environment.yaml 
```

Relevant error returned

``` bash
The conflict is caused by:
    The user requested sapien==2.2.1
    mani-skill 3.0.0b9 depends on sapien==3.0.0.b1

To fix this you could try to:
1. loosen the range of package versions you\'ve specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

Need to:
- Update `sapien` dependency within tdmpc2 baseline

PR #527, opened 2024-08-26
[BugFix] Bump sapien dependency in tdmpc2 baseline for mani_skill update

chenyenru:
Update `environment.yaml` for tdmpc2 baseline. See #526


StoneT2000:
Thanks!

PR #531, opened 2024-08-28
Added behavior cloning baselines

arnavg115:
Added the behavior cloning baseline for both state and rgbd observations.

### DIFF
--- a/examples/baselines/bc/bc.py
+++ b/examples/baselines/bc/bc.py
@@ -1,0 +1,408 @@
+import datetime
+import os
+import random
+import uuid
+from dataclasses import asdict, dataclass
+from typing import Optional
+
+import gymnasium as gym
+import h5py
+import numpy as np
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.optim as optim
+import tyro
+import wandb
+from mani_skill.utils.io_utils import load_json
+from mani_skill.utils.wrappers.gymnasium import CPUGymWrapper
+from mani_skill.utils.wrappers.record import RecordEpisode
+from torch.utils.data import DataLoader, Dataset
+from torch.utils.data.sampler import BatchSampler, RandomSampler
+from mani_skill.vector.wrappers.gymnasium import ManiSkillVectorEnv
+from tqdm import tqdm
+
+
+@dataclass
+class Config:
+    # wandb project name
+    project: str = "maniskill"
+    # wandb group name
+    group: str = "maniskill"
+    # training dataset and evaluation environment
+    env: str = "PickCube-v1"
+    # total gradient updates during training
+    max_timesteps: int = int(1e6)
+    # training batch size
+    batch_size: int = 128
+    # maximum possible trajectory length
+    max_traj_len: int = 100
+    # evaluation frequency, will evaluate eval_freq training steps
+    eval_freq: int = int(500)
+    # record videos
+    video: bool = False
+    # video save path
+    video_path = "./videos"
+    # training random seed
+    seed: int = 42
+    # training device
+    device: str = "cuda"
+    # learning rate
+    lr = 3e-4
+    # number of parallel eval envs
+    num_envs = 50
+    # dataset directory
+    demo_path: str = ""
+    # log to wandb
+    wandb: bool = True
+    # normalize states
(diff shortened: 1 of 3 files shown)


examples/baselines/bc/bc.py
```diff
@@ -0,0 +1,408 @@
...
+        device=device,
+        load_count=args.load_count,
+        normalize_states=args.normalize_states,
+    )
+    stats = ds.get_state_stats()
+
+    if args.video:
+       os.makedirs(os.path.join(log_path, args.video_path))
+    gpu = args.sim_backend == "gpu"
+    envs = make_eval_envs(args.env, args.num_envs, stats, control_mode,gpu)
```
StoneT2000: we should have a CLI argument to modify the max episode steps number to be higher from the command line. The code assumes 100 at the moment? You can replace max_traj_len with max_episode_steps as an argument. This should then also be passed to the environment creation tool.

examples/baselines/bc/bc_rgbd.py
```diff
@@ -0,0 +1,510 @@
...
+        load_count=args.load_count,
+    )
+
+    gpu = args.sim_backend == "gpu"
+
+    if args.video:
+        os.makedirs(os.path.join(log_path, args.video_path))
+
+
+    envs = make_eval_envs(args.env, args.num_envs, control_mode, gpu)
```
StoneT2000: I recommend sticking to using the flatteng RGBD observation wrapper as used in the PPO rgb code instead of making all the extra process obs functions.

examples/baselines/bc/bc_rgbd.py
```diff
@@ -0,0 +1,510 @@
...
+                        fin_info = fin_info[mask]
+                        fin_info = {
+                            k: np.array([dic[k] for dic in fin_info]) for k in fin_info[0]
+                        }
+                        el_steps.append(fin_info["elapsed_steps"])
+                        if "success" in fin_info:
+                            successes.append(fin_info["success"])
+                    returns.append(rewards[mask])
+
+            log_dict["charts/eval_elapsed_steps"] = np.concatenate(el_steps).mean()
```
StoneT2000: for logged data, can you try to be as consistent as possible with the diffusion policy baseline?

Namely there should be

eval/success_once
eval/success_at_end
eval/episode_len
eval/return

examples/baselines/bc/bc_rgbd.py
```diff
@@ -0,0 +1,510 @@
...
+
+        if i % args.eval_freq == 0:
+            successes = []
+            el_steps = []
+            returns = []
+            success_once = np.zeros((args.num_envs,))
+            rewards = np.zeros((args.num_envs,))
+            obs, _ = envs.reset()
+
+            for _ in range(100):
```
StoneT2000: 100 is hardcoded here. should be max episode steps

examples/baselines/bc/bc_rgbd.py
```diff
@@ -0,0 +1,510 @@
...
+    if args.wandb:
+        wandb_init(asdict(args))
+
+    ds = ManiSkillDataset(
+        args.demo_path,
+        device=device,
+        load_count=args.load_count,
+    )
+
+    gpu = args.sim_backend == "gpu"
```
StoneT2000: rename gpu to `gpu_sim`. `gpu` alone is a bit ambigious

examples/baselines/bc/bc_rgbd.py
```diff
@@ -0,0 +1,510 @@
...
+    return nn.Sequential(*module_list)
+
+def wandb_init(config: dict) -> None:
+    wandb.init(
+        config=config,
+        project=config["project"],
+        group=config["group"],
+        name=log_name.replace(os.path.sep, "__"),
+        id=str(uuid.uuid4()),
+        tags=["BC_RGBD"]
```
StoneT2000: for consistency lets name the tag "bc" for both bc rgbd and bc state, and also label group to be "BehaviorCloning". Default project name in the tyro args should be ManiSkill.

moreover why is there a ID generated here? We don't need to use uuid.

StoneT2000 (changes requested):

PR #535, opened 2024-08-29
Fix typo in params SimConfig

alexcbb:
The parameter for SimConfig was out of date


StoneT2000:
I don't follow. The current code is correct using the latest pypi/git commit version of this repo. Everything was renamed from `cfg` to `config` for consistency.

PR #596, opened 2024-10-02
Update minimal_template.py

DhruvKadam-git:
1. reset Method: Initializes the environment to a starting state.

2. step Method: Applies an action to the environment and returns  the observation, reward, done flag, and additional info.

3. render Method: Provides a placeholder for rendering the environment, with a check for unsupported modes.

### DIFF
--- a/mani_skill/envs/minimal_template.py
+++ b/mani_skill/envs/minimal_template.py
@@ -60,3 +60,26 @@ class CustomEnv(BaseEnv):
     ):
         max_reward = 1.0
         return self.compute_dense_reward(obs=obs, action=action, info=info) / max_reward
+
+    def reset(self):
+        """Reset the environment to an initial state."""
+        self._load_scene({})
+        self._initialize_episode(torch.tensor([0]), {})
+        return self._get_obs_extra({})
+
+    def step(self, action: torch.Tensor):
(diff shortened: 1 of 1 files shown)


StoneT2000:
why was this PR made? What is the intended change here

DhruvKadam-git:
1. reset Method:
      Purpose: Resets the environment to its initial state.
      Benefit: This is essential for starting a new episode in 
      reinforcement learning. It ensures that the environment is in a 
      known state before the agent begins interacting with it.

2. step Method:
      Purpose: Applies an action to the environment and returns 
      the resulting observation, reward, done flag, and additional 
      info.
      Benefit: This method is crucial for the interaction loop in 
      reinforcement learning. It allows the agent to take actions and 
      receive feedback from the environment, which is necessary for 
      learning.

3. render Method:
       Purpose: Renders the environment for visualization.
       Benefit: Visualization is important for debugging and 
       understanding the agent's behavior. The render method 
       provides a way to see what the environment looks like, which 
       can help in diagnosing issues and improving the agent's 
       performance.

StoneT2000:
These don't explain the changes well. These methods already exist and are inherited. Moreover we do not recommend users to ever override these functions in the first place

PR #603, opened 2024-10-04
Update README.md

kushal34712:
fixed typo


StoneT2000:
Thanks, but some of these changes are not necessary.

I'm aware hacktoberfest is going on but please do not spam pull requests for very small changes like this. Please take a bug from the issues tab and see if you can address it, then make a PR.

PR #611, opened 2024-10-09
[Feature]: Add CPU multiprocessing for Panda Motionplanner's Trajecto…

chenyenru:
Uses an elementary method that lets each process in `num_procs` to generate `num_traj // num_procs` trajectory.

Discussion on next steps:
- I see that in motionplanner's solution folder, there are several `solve` methods defined for different tasks. For example, `solve` in [`pick_cube.py`](https://github.com/haosulab/ManiSkill/blob/7b84f94dfc07c3cfbb6df4b1c12d3515d4253278/mani_skill/examples/motionplanning/panda/solutions/pick_cube.py). And it seems to be using the `PandaArmMotionPlanningSolver` environment when solving. From my understanding, it might help if `solve` method is being moved to `PandaArmMotionPlanningSolver` and `PandaArmMotionPlanningSolver` is wrapped as a vectorized environment for multiprocessing. Would be greatly appreciated to know if this rationale sounds correct, thank you.

### DIFF
--- a/mani_skill/trajectory/merge_trajectory.py
+++ b/mani_skill/trajectory/merge_trajectory.py
@@ -1,60 +1,76 @@
 import argparse
 from pathlib import Path
-
 import h5py
+from mani_skill.utils.logging_utils import logger
 
 from mani_skill.utils.io_utils import dump_json, load_json
 
 
-def merge_h5(output_path: str, traj_paths, recompute_id=True):
-    print("Merge to", output_path)
+def merge_trajectories(output_path: str, traj_paths: list, recompute_id: bool = True):
+    """
+    Merges multiple JSON and H5 files into a single JSON and H5 file.
+
+    This function combines the contents of multiple JSON and H5 files. It keeps the first value for all keys
+    (other than "episodes") and logs a warning for any differences. The "episodes" from each JSON file are merged
+    into a single list, and the corresponding H5 data is copied to the output H5 file.
+
+    Args:
+        output_path (str): The path to the output H5 file. The corresponding JSON file will be saved with the same
+                           name but with a .json extension.
+        traj_paths (list): A list of paths to the input trajectory files (H5 files). The corresponding JSON files
+                           should have the same name but with a .json extension.
+        recompute_id (bool): If True, recompute the episode IDs to ensure they are unique. If False, keep the original
+                             episode IDs.
+
+    Raises:
+        AssertionError: If there is a conflict in the episode IDs when recompute_id is False.
+    """
+    logger.info(f"Merging {output_path}")
 
     merged_h5_file = h5py.File(output_path, "w")
     merged_json_path = output_path.replace(".h5", ".json")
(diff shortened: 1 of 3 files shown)


StoneT2000:
I understand the reasoning somewhat, but the solver is dependent on the task. The PandArmmotionPlanningSolver is meant to a general class for the panda arm for MP problems.

StoneT2000 (changes requested):
tried running the code. However it seems some information is missing from the jsons. Particularly "source_type" and "source_desc". Seems to occur after merging the jsons.

StoneT2000:
oh I see the bug, its the merge_trajectory.py merge_h5 code, it doesn't add those in. I think a fix would be to just copy any of the keys in the first json that aren't the keys being merged in the merge_h5 code. 

and if there is different data in the other parts of the json just log a warning that this key is different between the different jsons and we took the first one.

chenyenru:
Thanks for the notes! Would replay trajectory need these information too? If so, I can directly modify merge_h5 to include "source_type" and "source_desc"

Regarding PandArmmotionPlanningSolver, would it make sense if child classes are created for each task?

StoneT2000:
Child subclasses is another approach, either a function that calls the APIs of the solver or implements the solver. I personally don't see a good reason to have to use subclasses though in this case if one is only overriding one function.

chenyenru:
Thank you, I believe I misunderstood the concept of a vectorized environment. Thought since each process in the subenvironment would be parallelized, if we put the class inside, it would be parallelized.

Will work on the fix for merge_h5 according to https://github.com/haosulab/ManiSkill/pull/611#issuecomment-2405921335

mani_skill/trajectory/merge_trajectory.py
```diff
@@ -1,60 +1,76 @@
...
+                h5_file.copy(traj_id, merged_h5_file, new_traj_id)
+
+                # Copy json data
+                if recompute_id:
+                    ep["episode_id"] = cnt
+                merged_json_data["episodes"].append(ep)
+
+                cnt += 1
 
-    # Ignore commit info
```
chenyenru, line 57: Commit info is now preserved in the trajectory data because I thought it could be useful when the version of trajectory generation code is different (e.g., if someone made changes on panda motionplanner).

mani_skill/trajectory/merge_trajectory.py
```diff
@@ -1,60 +1,76 @@
...
-
-            # Copy json data
-            if recompute_id:
-                ep["episode_id"] = cnt
-            merged_json_data["episodes"].append(ep)
-
-            cnt += 1
-
-        h5_file.close()
+        logger.info(f"Merging{traj_path}")
```
chenyenru, line 37: Made this `logger.info` from `print`. Is there a recommended practice on when to use `logger.info` and when to use `print`?

StoneT2000: better to use logger, thanks for using the right tool here.

StoneT2000 (approved):

PR #640, opened 2024-10-20
Adding Action Chunking with Transformers (ACT) to baselines 

ywchoi02:
#617


StoneT2000:
before I review code, make sure to add a README.md similar to the other baselines, just use DP as a reference (how to setup the conda/mamba env, citation etc.)

examples/baselines/act/act/detr/backbone.py
```diff
@@ -0,0 +1,122 @@
...
+from collections import OrderedDict
+
+import torch
+import torch.nn.functional as F
+import torchvision
+from torch import nn
+from torchvision.models._utils import IntermediateLayerGetter
+from typing import Dict, List
+
+from ..utils import NestedTensor, is_main_process
```
StoneT2000: use absolute imports when possible, it is just the style choice this repo uses.

ywchoi02: I changed them to absolute imports, but I'm not sure if they are correct. please let me know if they need to be fixed. I also added a README file.

examples/baselines/act/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+Code for running the ACT algorithm based on ["Learning Fine-Grained Bimanual Manipulation with Low-Cost Hardware"](https://arxiv.org/pdf/2304.13705). It is adapted from the [original code](https://github.com/tonyzhaozh/act).
+
+## Installation
+
+To get started, we recommend using conda/mamba to create a new environment and install the dependencies
+
+```bash
+conda create -n act-ms python=3.9
+conda activate act-ms
```
StoneT2000, line 11: so the conda env act-ms is created and you do a local pip install. However a simple setup.py file is still missing, can you create that?

examples/baselines/act/act/detr/backbone.py
```diff
@@ -0,0 +1,121 @@
...
+from collections import OrderedDict
+
+import torch
+import torch.nn.functional as F
+import torchvision
+from torch import nn
+from torchvision.models._utils import IntermediateLayerGetter
+from typing import Dict, List
+
+from examples.baselines.act.act.utils import NestedTensor, is_main_process
```
StoneT2000: imports should be absolute and relative to act (which you pip install -e .)

examples/baselines/act/act/detr/position_encoding.py
```diff
@@ -0,0 +1,93 @@
+# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+"""
+Various positional encodings for the transformer.
+"""
+import math
+import torch
+from torch import nn
+
+from examples.baselines.act.act.utils import NestedTensor
```
StoneT2000: same issue as above

StoneT2000 (changes requested):

examples/baselines/act/README.md
ywchoi02: I created a simple setup.py.

examples/baselines/act/act/detr/backbone.py
ywchoi02: updated my code.

examples/baselines/act/act/detr/position_encoding.py
ywchoi02: updated my code.

StoneT2000:
Final thing probably, can you create an examples.sh script for users to run? Should contain 2 scripts for a few envs, 1 for demonstration loading/replaying and 1 for training. Maybe just 2 for state based and 2 exampls for RGBD that work okay is fine.

examples/baselines/act/train_rgbd.py
```diff
@@ -0,0 +1,607 @@
...
+        self.include_state = state
+        self.transforms = T.Compose(
+            [
+                T.Resize((224, 224), antialias=True),
+            ]
+        )  # resize the input image to be at least 224x224
+        new_obs = self.observation(self.base_env._init_raw_obs)
+        self.base_env.update_obs_space(new_obs)
+
+    def observation(self, observation: Dict):
```
StoneT2000, line 144: is there a reason why the maniskill provided flatten rgbd wrapper cant be used? save some code here.

StoneT2000 (approved):

StoneT2000:
just one comment and then we can merge before starting doing a larger sweep of experiments to upload to wandb

examples/baselines/act/train_rgbd.py
ywchoi02: 1. Visual Data Shape: 
    The visual data (rgb or rgbd) is of shape [batch, num_cams, C, H, W] instead of [batch, H, W, num_cams*C] as it makes it easier to  use as input to CNN backbone. The CNN backbone accepts input of size [batch, C, H, W] so we query the backbone for each camera image and concatenate them afterwards.

2. Visual Data Size: 
    The pre-trained models from torchvision.models expect input image to be at least 224x224, and the default visual data size is 128x128. I learned recently that you can change this with sensor_config param in gym.make but if this only changes the rgbd returned by the env and not the rgbd from the replayed trajectory dataset, we would still need to resize the rgbd from the replayed trajectory dataset, and there would be a discrepancy between the two rgbd. I'm not sure how much this discrepancy will affect the performance as I haven't tested them, but resizing it in the wrapper makes it consistent.

StoneT2000: Ok no worries then.

Issue #638, opened 2024-10-20
[Feature] Tasks to add


StoneT2000:
A list of tasks planned to be added to ManiSkill and eventually come with demos.

- [x] OpenCabinetDrawer
- [x] Multiple drawing related tasks
- [x] Clean table task (2 versions, robot with "sponge" attached" and one without probably)
- [x] Long horizon pyramid building tasks
- [x] More "household" type of tasks, which dynamically are not any more difficult than table-top but "look" more realistic and interesting

PR #643, opened 2024-10-21
[Feature] Added RGBD diffusion policy implementation as well as Draw Triangle and Draw SVG task

arnavg115:
DrawTriangle, DrawSVG: #638
Diff policy RGBD: #616


mani_skill/examples/motionplanning/panda_stick/run.py
```diff
@@ -0,0 +1,154 @@
...
+import numpy as np
+from mani_skill.examples.motionplanning.panda_stick.solutions import \
+    solveDrawTriangle
+from mani_skill.utils.wrappers.record import RecordEpisode
+from tqdm import tqdm
+
+MP_SOLUTIONS = {"DrawTriangle-v1": solveDrawTriangle}
+
+
+def parse_args(args=None):
```
StoneT2000, line 14: thanks for adding this file. 

we might want to consider having a more unified motion planning interface for all robots with some base classes and run scripts etc., this should be fine for now though!

mani_skill/utils/wrappers/flatten.py
```diff
@@ -50,7 +51,12 @@ def observation(self, observation: Dict):
         if self.include_rgb and not self.include_depth:
             ret["rgb"] = images
         elif self.include_rgb and self.include_depth:
-            ret["rgbd"] = images
+            # todo: might not be the best way or the right place to perform this
```
StoneT2000: remove the todo. I think this is fine. I normally myself wouldn't merge rgb and depth but this is just for compatibility with a lot of old code from MS2 that assumed "rgbd" as a key.

StoneT2000 (changes requested):

mani_skill/envs/tasks/drawing/draw_triangle.py
```diff
@@ -0,0 +1,330 @@
+import math
```
StoneT2000, line 1: image

Is it possible to make the outline 1. not transparent and 2. just be below the dots/ink? instead of over it. Looks a bit strange here

mani_skill/envs/tasks/drawing/draw_triangle.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+@register_env("DrawTriangle-v1", max_episode_steps=1000)
+class DrawTriangle(BaseEnv):
+    """
+    This is a simple environment demonstrating drawing simulation on a table with a robot arm. There are no success/rewards defined, users can use this code as a starting point
+    to create their own drawing type tasks.
+    """
+
+    MAX_DOTS = 1010
```
StoneT2000: reduce this down significantly, maybe to 300. and max episode steps should be down to 300

mani_skill/examples/motionplanning/panda_stick/solutions/draw_triangle.py
```diff
@@ -0,0 +1,40 @@
...
+
+def solve(env: PushCubeEnv, seed=None, debug=False, vis=False):
+    env.reset(seed=seed)
+    planner = PandaStickMotionPlanningSolver(
+        env,
+        debug=debug,
+        vis=vis,
+        base_pose=env.unwrapped.agent.robot.pose,
+        visualize_target_grasp_pose=vis,
+        print_env_info=False,
```
StoneT2000, line 15: can you lower the max allowed joint velocities? Right now the result drawings are a bit too fast (dots are too spread out).

maybe lets tune it so that the typical solve time is about the 200-250 steps mark

StoneT2000 (changes requested):

examples/baselines/diffusion_policy/diffusion_policy/make_env.py
```diff
@@ -1,12 +1,73 @@
...
+from gymnasium.wrappers.frame_stack import FrameStack as GymFrameStack
+from gymnasium.wrappers.frame_stack import LazyFrames
 from mani_skill.utils import gym_utils
+from mani_skill.utils.wrappers import CPUGymWrapper, FrameStack, RecordEpisode
 from mani_skill.vector.wrappers.gymnasium import ManiSkillVectorEnv
-from mani_skill.utils.wrappers import RecordEpisode, FrameStack, CPUGymWrapper
 
 
-def make_eval_envs(env_id, num_envs: int, sim_backend: str, env_kwargs: dict, other_kwargs: dict, video_dir: Optional[str] = None, wrappers: list[gym.Wrapper] = []):
+class DictFrameStack(GymFrameStack):
```
StoneT2000: can this not just inherit the maniskill frame stack wrapper?

I purposely didn't use the original framestack wrapper from gymnasium since it was not properly GPU parallelized.

StoneT2000 (changes requested):
Add a new baselines.sh file and update README to point to that file. Write a script to train for each of the tasks you tested, using the datasets generated by the script here: https://github.com/haosulab/ManiSkill/blob/main/scripts/data_generation/replay_for_il_baselines.sh

Also make sure to add a draw triangle replay trajectory script to the file above.

StoneT2000:
all the changes nearly look ready. I will make a few small edits on baselines.sh file to be closer to what we did to act baselines. The diffusion policy one i did before for state based results is kind of hard to read, it was meant to experiment with DP training on motion planning or RL generated datasets (but doing that needs tuning other hyparameters anyway).

Will also merge in #870 which is blocking me from easily testing all the baselines and uploading results to wandb since things take forever for the physx_cpu evals.

StoneT2000:
Merging to a separate branch for now so i make the final minor changes and merge. nice work!

PR #648, opened 2024-10-23
[Feature] Support the Robocasa scene dataset

StoneT2000:
## Features
- [x] Import and parallelize loading the scenes
- [x] Load the random objects in the scene
- [ ] Rendering tricks to fake water flow + turning on stoves
- [x] Align camera parameters

This pr does not include task definitions (with success conditions and what not). It is best used as a starting point (inherit the RobocasaKitchenEnv class) to create your own tasks.

## Other
- fixes a bug where non static objects that had their initial poses set would not be initialized there on CPU sim but would be on the GPU sim.
- shadows are enabled by default on demoscripts since they look nicer and demos are not meant for performance testing.
- introduces repeated plane textures similar to the texrepeat property mujoco has. 
- fixes a number of small bugs for loading MJCFs.
- adds a better worded message when user tries to set pose of static objects on the GPU sim
- support "none" as a robot uid for just pure exploration and no actions/controllers.


dreamdev123:
What is your problem?

PR #655, opened 2024-10-25
[BugFix] Improve GPU sim speed and new docs on initial pose handling

StoneT2000:


StoneT2000:
We discovered a few important things that can easily slow down the GPU simulation (and even break it!), primarily around setting initial poses of builders before building robots/actors/articulations. The latest commit and eventually future releases have new details on how to make GPU simulation faster and importantly accurate. Existing envs you make are not affected really but you may get some warnings from our code about not setting initial poses.

In particular, we strongly recommend the following when building your own tasks
- override the _load_agent function to set the robot's initial pose (you cannot do it anywhere else). You should pick a pose that isn't too big in values but also is unlikely to collide with other objects if it was placed there. This has caused bugs when e.g. the initial pose of a robot and cabinet were both at 0, and GPU sim uses initial poses to initialize contact buffers and other data. The intersections are bad for GPU sim especially for articulated objects.
```
import sapien
class YourEnv(BaseEnv):
    def _load_agent(self, options: dict):
        super()._load_agent(options, sapien.Pose(p=[0, 0, 1]))
```
- ensure to specify an initial pose in your actor / articulation builders. Again following the same general rule of thumb to place objects such that they don't intersect.
```
    builder.initial_pose = sapien.Pose(p=[0, 1, 0]) # for example
    builder.build()
```

For envs specifically with complex articulations and more than 1, these fixes are critical to ensuring less bugs occur and for faster simulation. Docs for these recommendations have been updated on the custom tasks tutorial: https://maniskill.readthedocs.io/en/latest/user_guide/tutorials/custom_tasks/intro.html#loading

PR #656, opened 2024-10-26
[BugFix]: improve perf for batched_rng

arth-shukla:
- [x] set batchrng by env_idx when resetting
- [x] double-set seeds only if reconfiguring
- [x] allow set batchrng with env_idx
- [x] allow call batchrng with env_idx
- [x] use batchrng with env_idx for table scene builder robot init
- [x] enchanced_determinism option

<p>
<em>grey: b10, purple: latest::enhanced_determinism=<b>off</b>, blue: latest::enhanced_determinism=<b>on</b></em>
image
</p>


StoneT2000:
nice work

PR #670, opened 2024-10-31
[Fix] Update replay_trajectory.py

PartyPenguin:
episode_seed is now a numpy array so the current version is not working.  To fix it we need to just get the first element of the array.


StoneT2000:
Thanks for the bug spot. The actual fix is in the record episode wrapper #671 but this can help as well (given some users may have generated some data already).

StoneT2000:
can you modify this to check if the item is a list, and if so parse it as an int and assert there is only one element? each trajectory should only ever be associated with one seed. We recently aligned CPU/GPU sim seeds by having a batch of seeds for each parallel env which caused this bug.

PR #698, opened 2024-11-15
[Feature] SAC RGB(D) Baseline and improved memory efficiency

StoneT2000:


StoneT2000:
Tests on new SAC state and SAC RGBD baselines on PickCube-v1

black line is state based, green is 64x64 RGB images, purple is 128x128 RGB images. Note 128x128 (the default camera size in most tasks) runs very slowly due to very slow update times. Eventually will make a sac_fast version based on leanrl. 
image

PR #700, opened 2024-11-16
Fix flatten_state_dict with empty dictionary

younik:
[torch.Tensor.size](https://pytorch.org/docs/stable/generated/torch.Tensor.size.html) returns as tensor.Size object when the dimension is not specified, making this check failing:
https://github.com/haosulab/ManiSkill/blob/b9c17c8e9af4b831178a55151d5434c99de448cc/mani_skill/utils/common.py#L200-L201


StoneT2000:
Thanks!

StoneT2000 (approved):

PR #761, opened 2024-12-22
added property to base env for benchmarking

rrzhang139:
After genesis base env setup, sapien env was not working due to the missing `fixed_trajectory` property.


StoneT2000:
Which code wasn't working? Do you mean running gpu_sim.py on other environments?

If so the correct fix really should be to check if fixed trajectory property exists (and if not don't run that part of the benchmark).

StoneT2000:
closed in favor of #762. Thanks for the bug spot

PR #774, opened 2024-12-26
PullDrawer-v1 Env with reward function and motion planner

Viswesh-N:
Added new env: PullDrawer-v1 along with the dense reward and motion planner


mani_skill/envs/tasks/tabletop/pull_drawer.py
```diff
@@ -0,0 +1,360 @@
...
+from mani_skill.utils import sapien_utils
+from mani_skill.utils.registration import register_env
+from mani_skill.utils.scene_builder.table import TableSceneBuilder
+from mani_skill.utils.structs.types import SimConfig, GPUMemoryConfig
+from mani_skill.sensors.camera import CameraConfig
+from mani_skill.utils.structs import Pose
+from mani_skill.utils.building import actors
+
+@register_env("PullDrawer-v1", max_episode_steps=200)
+class PullDrawerEnv(BaseEnv):
```
StoneT2000, line 16: we recently updated how we document tasks/environments. It is nearly completely automatically generated now. Can you follow the documentation style of e.g. PushCube? Moreover can you save an example video with ray tracing of this task and put it in the folder with the others?

StoneT2000 (changes requested):
Main things need example video task, document it as a python docstring (see other tasks), and also can you update the code generation scripts in scripts/data_generation with the new motion planner code?

And for PPO can you give an example script that works? I will run it on my 4090 to record to wandb.

PR #783, opened 2024-12-31
Fix: vector-gymnasium-wrapper -- fixing 'call' method -- getattr(self.env...) …

imishani:
Fixing tiny bug in `ManiSkill/mani_skill/vector/wrappers/gymnasium.py`, in function `call`, line 162: 
from 
`function = getattr(self.env, name)`
to 
`function = getattr(self._env, name)`


StoneT2000 (approved):
thanks!

PR #820, opened 2025-02-03
[Feature] Support XArm6 with robot iq gripper

StoneT2000:


StoneT2000:
@dwaitbhatt what is a typical "rest" position of the xarm6?

One small complication is in most environments the starting pose of the end-effector is different to the panda arm, so by default RL will not work for most of those table-top franka environments because we tune the max episode steps to be low but sufficient enough for the panda arm to solve the task.

If there is a typical rest pose that is a bit closer to the panda one maybe we make that the default for the table top scene. This way then I can just add the xarm6_robotiq as a supported robot for a few tested environments that should work.

StoneT2000:
Actually let me make this just a separate discussion on GitHub here https://github.com/haosulab/ManiSkill/discussions/823. I will merge the tested code first and remove xarm6 from one of the default options permitted in PickCube (needs more tuning to work robustly).

PR #828, opened 2025-02-05
Fixed final_info having wrong elapsed_steps value after reset and final_observation being wrong on CUDA

AlexandreBrown:
# Description
- Because `BaseEnv::get_info` does not always clone the `elapsed_steps`, the dict still pointed to the same tensor therefore the `env.reset` would change the `elapsed_steps` stored (even if it's a new dict, the value was still the same tensor) which would invalidate the `final_info` data.
- My change is local to the vect env to prevent major change across the code base, an alternative would be to make `BaseEnv::get_info` always clone the `elapsed_steps` but since there is currently an `if` statement that prevents this I assume this is for a specific reason, so I decided to not change `BaseEnv::get_info`.  
```python
def get_info(self) -> dict:
        """
        Get info about the current environment state, include elapsed steps and evaluation information
        """
        info = dict(
            elapsed_steps=self.elapsed_steps
            if not self.gpu_sim_enabled
            else self._elapsed_steps.clone()
        )
        info.update(self.evaluate())
        return info
```  
[source](https://github.com/haosulab/ManiSkill/blob/a1bc12628ebc6ac48c2150b6bd79e36908055fb3/mani_skill/envs/sapien_env.py#L1016C5-L1026C20)

# TLDR:   
If we take for example a task with auto_reset and a max traj length of 50 (eg: Push Cube).  
## Before This PR    
```
obs, info = env.step(...) # info["elapsed_steps"] would be 49
obs, info = env.step(...) # info["elapsed_steps"] would be 0 (since we use auto_reset)
print(info["final_info"]["elapsed_steps"]) # Gives 0 (wrong!)
```
## After This PR    
```
obs, info = env.step(...) # info["elapsed_steps"] would be 49
obs, info = env.step(...) # info["elapsed_steps"] would be 0 (since we use auto_reset)
print(info["final_info"]["elapsed_steps"]) # Gives 50 (the true elapsed_steps of the true last observation) 
```

### DIFF
--- a/mani_skill/vector/wrappers/gymnasium.py
+++ b/mani_skill/vector/wrappers/gymnasium.py
@@ -4,10 +4,10 @@ from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
 
 import gymnasium as gym
 import torch
-import copy
 from gymnasium.vector import VectorEnv
 
 from mani_skill.utils.structs.types import Array
+from mani_skill.utils.common import torch_clone_dict
 
 if TYPE_CHECKING:
     from mani_skill.envs.sapien_env import BaseEnv
@@ -142,9 +142,9 @@ class ManiSkillVectorEnv(VectorEnv):
         dones = torch.logical_or(terminations, truncations)
 
         if dones.any() and self.auto_reset:
-            final_obs = copy.deepcopy(obs)
+            final_obs = torch_clone_dict(obs)
             env_idx = torch.arange(0, self.num_envs, device=self.device)[dones]
-            final_info = copy.deepcopy(infos)
+            final_info = torch_clone_dict(infos)
             obs, infos = self.reset(options=dict(env_idx=env_idx))
             # gymnasium calls it final observation but it really is just o_{t+1} or the true next observation
             infos["final_observation"] = final_obs
--- a/mani_skill/vector/wrappers/gymnasium.py
+++ b/mani_skill/vector/wrappers/gymnasium.py
@@ -4,6 +4,7 @@ from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
 
 import gymnasium as gym
 import torch
+import copy
 from gymnasium.vector import VectorEnv
 
 from mani_skill.utils.structs.types import Array
@@ -143,8 +144,8 @@ class ManiSkillVectorEnv(VectorEnv):
         if dones.any() and self.auto_reset:
             final_obs = obs
             env_idx = torch.arange(0, self.num_envs, device=self.device)[dones]
+            infos["final_info"] = copy.deepcopy(infos)
             obs, _ = self.reset(options=dict(env_idx=env_idx))
-            infos["final_info"] = infos.copy()
             # gymnasium calls it final observation but it really is just o_{t+1} or the true next observation
             infos["final_observation"] = final_obs
             # NOTE (stao): that adding masks like below is a bit redundant and not necessary
--- a/mani_skill/vector/wrappers/gymnasium.py
+++ b/mani_skill/vector/wrappers/gymnasium.py
@@ -144,10 +144,11 @@ class ManiSkillVectorEnv(VectorEnv):
         if dones.any() and self.auto_reset:
             final_obs = obs
             env_idx = torch.arange(0, self.num_envs, device=self.device)[dones]
-            infos["final_info"] = copy.deepcopy(infos)
-            obs, _ = self.reset(options=dict(env_idx=env_idx))
+            final_info = copy.deepcopy(infos)
+            obs, infos = self.reset(options=dict(env_idx=env_idx))
             # gymnasium calls it final observation but it really is just o_{t+1} or the true next observation
             infos["final_observation"] = final_obs
+            infos["final_info"] = final_info
             # NOTE (stao): that adding masks like below is a bit redundant and not necessary
             # but this is to follow the standard gymnasium API
             infos["_final_info"] = dones
--- a/mani_skill/vector/wrappers/gymnasium.py
+++ b/mani_skill/vector/wrappers/gymnasium.py
@@ -142,7 +142,7 @@ class ManiSkillVectorEnv(VectorEnv):
         dones = torch.logical_or(terminations, truncations)
 
         if dones.any() and self.auto_reset:
-            final_obs = obs
+            final_obs = copy.deepcopy(obs)
             env_idx = torch.arange(0, self.num_envs, device=self.device)[dones]
             final_info = copy.deepcopy(infos)
             obs, infos = self.reset(options=dict(env_idx=env_idx))
--- a/mani_skill/utils/common.py
+++ b/mani_skill/utils/common.py
@@ -16,6 +16,23 @@ from mani_skill.utils.structs.types import Array, Device
 # Utilities for working with tensors, numpy arrays, and batched data
 # -------------------------------------------------------------------------- #
 
+def torch_clone_dict(data: dict) -> dict:
+    """
+    Recursively clones all torch tensors in a dictionary.
+    If the input was a torch tensor, it will return a clone of the tensor.
+    """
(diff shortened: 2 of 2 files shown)


AlexandreBrown:
Hi @StoneT2000 , would love some feedback on why that if exists in `BaseEnv::get_info`, I assumed there is a good reason for it so I didn't touch it and did the fix more locally.

StoneT2000:
You are correct, the final elapsed steps value is being overwritten

That being said is it possible to avoid a deep copy here?

Or is the overhead very small

AlexandreBrown:
@StoneT2000 The overhead is very small since the biggest tensor that is being copied is the `elapsed_steps` tensor which is of dimension `(num_envs,)` and of dtype int.  
The deep copy would not be needed if `BaseEnv::get_info` would always `clone()` the `elapsed_steps` tensor but it seems like there is an `if` statement that prevents this from happening on CPU env, any reason why?  

**Update:** 
After doing more tests I found that another issue could happen on cuda.  
Basically, we had this  :    
## Before Second Commit of This PR (when using CUDA/CPU)   
```python
..., infos = env.step() # infos['elapsed_steps'] is 50

if auto reset:
    obs, _ = env.reset() # infos from reset is unused!

return obs ,..., infos # infos['elapsed_steps'] will be 0 if we use cpu since no clone() is done for elapsed_steps when using CPU but will be 50 for cuda since a clone() is done so only the result of `env.reset` would have the `infos` with the updated `elapsed_steps`.
```
## After Second Commit of This PR (when using CUDA/CPU)   
I have added a commit that makes it so now the code works regardless of CPU/CUDA, we always get the expected result since we now use the infos dict returned by the auto_reset :   
```python
..., infos = env.step()

if auto reset:
    obs, infos = env.reset() # infos from reset is used, elapsed_steps will be 0 regardless of CPU/CUDA since it's the lastest value

return obs ,..., infos # infos['elapsed_steps'] will be 0  and infos['final_info']['elapsed_steps'] will be 50 which is the correct output
```

AlexandreBrown:
**Update 2 :**  
After some more testing I actually discover a third issue.  
The `final_obs` is correct when using `cpu` but when using `cuda`, the following code:  
```python
final_obs = obs
env_idx = torch.arange(0, self.num_envs, device=self.device)[dones]
final_info = copy.deepcopy(infos)
obs, infos = self.reset(options=dict(env_idx=env_idx))
# gymnasium calls it final observation but it really is just o_{t+1} or the true next observation
infos["final_observation"] = final_obs 
```
Will actually result in `final_observation` being equal to `obs` (the reset obs!).  
I pushed another commit to fix this so that the CUDA behavior is the same as CPU.  
I have manually verified the images stored in `final_observation` and I can confirm that the fix resolves this issue.

StoneT2000:
😮 

i think ill check if this is the case and see how RL performance in our baselines change. I somehow did not notice this was an issue

this may be because PPO is trained with partial resets typically so the final observation is not really used? Not sure

StoneT2000:
@AlexandreBrown Actually i cannot reproduce this

```
> /home/stao/work/research/maniskill/ManiSkill/mani_skill/vector/wrappers/gymnasium.py(150)step()
    149             import ipdb; ipdb.set_trace()
--> 150             infos["final_observation"] = final_obs
    151             # NOTE (stao): that adding masks like below is a bit redundant and not necessary

ipdb> final_obs.keys()
dict_keys(['state', 'rgb'])
ipdb> final_obs["rgb"].mean()
*** RuntimeError: mean(): could not infer output dtype. Input dtype must be either a floating point or complex dtype. Got: Byte
ipdb> final_obs["rgb"].float().mean()
tensor(123.3990, device='cuda:0')
ipdb> obs["rgb"].float().mean()
tensor(123.3132, device='cuda:0')
ipdb> 
```

StoneT2000:
I can confirm the issue for the elapsed steps number being wrong on CPU envs, which is due to not cloning the tensor. You can just add a .clone there

Although in the future I may wish to investigate whether we should have so many .clone or new tensor creations (it would be better to fill an existing allocated tensor but need performance checking).

AlexandreBrown:
> @AlexandreBrown Actually i cannot reproduce this

Can you please check on cuda? The issue for `final_observation` was only on cuda.

StoneT2000:
This is on the sim_backend="physx_cuda". (the data is on device cuda as well).

This is an issue of the MS3 wrapper, not your own?

AlexandreBrown:
Yes this is with the MS3 wrapper `ManiSkillVectorEnv`.  
It seems like only the visual data is the same (when we do not deep copy the `final_obs`) :    
image  
image  
Tested using :  
`sim_backend = "gpu"`  
`render_backend = "gpu"`  
Note that I have 2 env instances (one for train and one for eval).

StoneT2000:
Ok I think I am able to reproduce this issue now. It seems image data is not being copied but state data is as you say. 

Luckily it seems the RL baselines that use the FlattenRGBDObservationWrapper do not have this issue it would appear (I think the torch.cat operation allocates a new tensor)

mani_skill/vector/wrappers/gymnasium.py
```diff
@@ -141,12 +142,13 @@ def step(
         dones = torch.logical_or(terminations, truncations)
 
         if dones.any() and self.auto_reset:
-            final_obs = obs
+            final_obs = copy.deepcopy(obs)
             env_idx = torch.arange(0, self.num_envs, device=self.device)[dones]
-            obs, _ = self.reset(options=dict(env_idx=env_idx))
-            infos["final_info"] = infos.copy()
+            final_info = copy.deepcopy(infos)
```
StoneT2000: we should replace the deepcopy operation with one designed for torch to prefer using torch.clone

maybe a util called torch_clone_dict in the utils/common.py file?

mani_skill/vector/wrappers/gymnasium.py
```diff
@@ -141,12 +142,13 @@ def step(
...
+            obs, infos = self.reset(options=dict(env_idx=env_idx))
```
StoneT2000, line 148: Is gymnasium infos supposed to be done like this?

If i recall correctly generally on the first environment reset there is no useful infos, just reset info details. Althoughin the ManiSkill case the reset info is using information from the get_info function.

StoneT2000 (changes requested):

mani_skill/vector/wrappers/gymnasium.py
AlexandreBrown: `infos` is a dict not a tensor, are you suggesting that we add a util that iterates over the dict and performs a recursive search and `.clone` if `isinstance(x, torch.Tenosr)` ?

mani_skill/vector/wrappers/gymnasium.py
AlexandreBrown: I am assuming that the existing `self.reset` does what we want, since this `obs, infos = self.reset(options=dict(env_idx=env_idx))` is called inside `step` and so the env will already have performed its initial `reset` ex:  
```python
obs, _ = ms3_env.reset(...)

for _ in range(...):
   obs, rew, terminations, truncations, infos = ms3_env.step() # Reset was already called before we call step
```

mani_skill/vector/wrappers/gymnasium.py
StoneT2000: yes

info is a dict but may have tensor elements

AlexandreBrown: Sorry for the delay, just pushed your suggestion.  
I made sure to handle the fact that obs can be a dict or tensor (eg: if using only state obs_mode it would be a tensor).

AlexandreBrown:
@StoneT2000 Should be good to merge now, let me know if there is anything missing.  
Once this is merged, then https://github.com/haosulab/ManiSkill/pull/832 can be merged/review

StoneT2000 (approved):

PR #876, opened 2025-02-22
[Feature] Added RGB DP baseline and DrawTriangle/SVG task

StoneT2000:
- Adds in @arnavg115 work on a RGB Diffusion Policy Baseline and cleans the script up and improves documentation around IL in general
- Make DrawTriangle a bit more strict but still solvable
- New and improved FrameStack wrapper that works on both CPU and GPU sims and handles dictionary observations. Seems to be as fast as the gymnasium FrameStack wrapper that uses LazyFrames.
- Removes the old generate.sh file for motionplanning demo generation and moves it with the other data generation scripts. motionplanning runner code for panda updated to run both panda stick and panda gripper solutions.
- Fix bug with to_cpu_tensor not actually moving all tensors to the CPU if input was a dict
- Update FlattenRGBDObservation wrapper to always put depth and rgb data separately by default. There isn't a good reason to put them together as these two modalities will always be have to handled differently regardless due to different normalizations.


mani_skill/utils/wrappers/frame_stack.py
```diff
@@ -119,7 +46,24 @@ def base_env(self) -> BaseEnv:
...
     def observation(self, observation):
-        return torch.stack(list(self.frames)).transpose(0, 1)
+        assert len(self.frames) == self.num_stack, (len(self.frames), self.num_stack)
+        if self.use_dict:
+            return {
+                k: torch.stack([x[k] for x in self.frames], dim=0).transpose(0, 1)
+                for k in self.observation_space.keys()
+            }
+        else:
+            return torch.stack(list(self.frames)).transpose(0, 1)
```
StoneT2000, line 56: is torch.roll and ring buffer faster here?

PR #918, opened 2025-03-09
[BugFix]: ycb checksum

arth-shukla:
ycb checksum needed to be updated

https://github.com/arth-shukla/mshab/issues/10


StoneT2000:
we should probably git rid of these checksums, and version control with git lfs on HF or something in the future

Issue #650, opened 2024-10-24
[Question] GPU mem not available when trying the gpu_sim example


kevinqyh0827:
Hi, I was trying running the "mani_skill.examples.benchmarking.gpu_sim" example from the docementation website. I got this error below:
![image](https://github.com/user-attachments/assets/a5bc63bf-a3ea-41f0-a8f1-79ef5cfcab04)

I add 2 lines to print the cpu_mem and gpu_mem. It shows the "gpu_mem" is "NonType". I wondering if it's because my gpu is not working or other issues. 

I was running ManiSkill in docker and my server is Ubuntu 22.04, GPU is RTX 3090. The output info of "nvidia-smi" inside the docker container is:
![image](https://github.com/user-attachments/assets/ca980759-aec2-4244-9cb1-df0812473b2e)
If it is the Ubuntu version issue, currently I am done since I do not have the permission to downgrade it.

Thanks so much!

StoneT2000:
Is your docker correctly setup for NVIDIA drivers? See https://maniskill.readthedocs.io/en/latest/user_guide/getting_started/docker.html for nvidia containers

kevinqyh0827:
Hi, thanks for your reply. I noticed the doc suggested to install "nvidia-docker v2". But search results and official instructions both indicate this wrapper is no longer supported. They sugggest to use "nvidia-container-toolkit". Here is the "nvidia-container-toolkit" and other tools installed on the server:
![image](https://github.com/user-attachments/assets/72ecffc4-42a0-442c-abfe-5dbc9c90b7a0)

I think it is correctly setup since I have finished other tasks inside a docker container requiring access to GPUs. Is this version suitable for Maniskill? Is there any other way or other example code in this repo to test this issue? 

Thanks!

StoneT2000:
unfortunately I am not really sure here why nvidia-smi is not showing the GPU usage. the thing is you are most likely using ti since the gpu sim script did run (just couldn't report GPU usage). the 22k fps seems about right with the default settings

kevinqyh0827:
Hi, thanks for your suggestions. Sorry for the misleading nvidia-smi image. I wanted to show that the GPU driver is correct in the initial problem description. Here is the usage of GPUs when running the "gpu_sim" code:
![image](https://github.com/user-attachments/assets/a040ffb9-428c-4ca6-95a2-b94339fb7378)

[...]

PR #924, opened 2025-03-11
fix gpu sim demo, mem type is None

SajjadPSavoji:
When using the docker container, I noticed that the GPU mem used in gpu sim demo was 0. There is a fix already suggested for that in the following issue. 
https://github.com/haosulab/ManiSkill/issues/650
by adding `--pid host` the gpus are utilized from inside the container.
Lets add this line to the docs to pervent others from having trouble with docker.
[Solution](https://github.com/haosulab/ManiSkill/issues/650#issuecomment-2434812353)


StoneT2000 (approved):
thanks!

PR #936, opened 2025-03-15
[Feature] Support motionplanning solutions for xarm6_robotiq

dwaitbhatt:
- Adding xarm6 motionplanning solutions for PickCube, PushCube, StackCube, and PlugCharger. Attaching sample videos for reference.
- To support xarm6 motionplanning, we need new .convex.stl assets. Also attaching the updated assets archive with which I was able to successfully execute motionplanning with xarm6_robotiq. 
[ManiSkill-XArm6-0.1.1.zip](https://github.com/user-attachments/files/19260568/ManiSkill-XArm6-0.1.1.zip)

https://github.com/user-attachments/assets/83bf4f1c-fefc-4a12-b9c1-03df6597a74d

https://github.com/user-attachments/assets/a0460849-6fd8-4506-b438-ae7a8fde19c7

https://github.com/user-attachments/assets/2b0be547-5dab-4545-8675-5ef6ef27e4b6

https://github.com/user-attachments/assets/adc42dda-6272-4a47-8947-5eb9c12d576e

### DIFF
--- a/mani_skill/agents/robots/xarm6/xarm6_robotiq.py
+++ b/mani_skill/agents/robots/xarm6/xarm6_robotiq.py
@@ -403,6 +403,18 @@ class XArm6Robotiq(BaseAgent):
         )
         return torch.logical_and(lflag, rflag)
 
+    @staticmethod
+    def build_grasp_pose(approaching, closing, center):
+        """Build a grasp pose ()."""
+        assert np.abs(1 - np.linalg.norm(approaching)) < 1e-3
+        assert np.abs(1 - np.linalg.norm(closing)) < 1e-3
+        assert np.abs(approaching @ closing) <= 1e-3
+        ortho = np.cross(closing, approaching)
+        T = np.eye(4)
+        T[:3, :3] = np.stack([ortho, closing, approaching], axis=1)
+        T[:3, 3] = center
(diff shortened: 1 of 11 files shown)


dwaitbhatt:
@StoneT2000 
Some notes/lessons that might be useful for supporting motionplanning with other new robots: 
- For some reason planning with screw fails a lot with xarm6 (but not with panda). For such cases, RRT works well.
- The default RRT variant is RRTConnect, which returns feasible but very weird paths for xarm6 (but not for panda). The optimal variant RRTStar is a lifesaver here to get sensible paths.
- I found the RRTStar hyperparams `rrt_range=0.0` (automatically figure out suitable range) and `planning_time=1` to work well for xarm6, might have to tune these for other robots.

StoneT2000 (approved):
thanks this looks great!

For a sanity check could you try running diffusion policy (probably easiest to setup imitation learning baseline) on the PickCube task to confirm the generated data is learnable?

you will need to replay the data (copy the script here and modify for your data, just change path perhaps: https://github.com/haosulab/ManiSkill/blob/main/scripts/data_generation/replay_for_il_baselines.sh)

and then do training

StoneT2000 (approved):

PR #976, opened 2025-04-02
Adding missing `.cpu()` calls in stack_cube.py, peg_insertion_side.py

jstmn:
Fixes a bug where .cpu() was missing in two files, causing an error:

```
Cannot find valid solution because of an error in motion planning solution: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
```


StoneT2000:
Thanks! Are you trying to collect data with the physx_cuda backend?

Just noting that we don't really support that at the moment (really just waiting on adding some batched motion planning library before doing that).

jstmn:
Hey! yes I am. Good to know about the cpu option. I'll use that from now on, it seems to be working consistently. 

Should we still merge these fixes anyway?

Issue #958, opened 2025-03-23
Question about demo_manual_control.py script (In ManiSkill/mani_skill/examples/demo_manual_control.py)


dongyingyibadao:
Hi! I’ve been analyzing the script “demo_manual_control.py” to understand how the robot in Maniskill can be controlled via keyboard. My plan is to create an interface for other devices, such as joysticks, etc. However, when I run the script and press 'w' to move the robot forward, I notice that not only does the robot move forward, but its mechanical arm also moves. Why is the motion of the arm not decoupled from the robot's movement?

Another issue arises in this part of the code:

```python
if has_base:
    if key == "w":  # forward
        base_action[0] = 1
    elif key == "s":  # backward
        base_action[0] = -1
    elif key == "q":  # rotate counterclockwise
        base_action[2] = 1
    elif key == "e":  # rotate clockwise
        base_action[2] = -1
    elif key == "z":  # lift
        body_action[2] = 1
    elif key == "x":  # lower
        body_action[2] = -1
    elif key == "v":  # rotate head left
        body_action[0] = 1
    elif key == "b":  # rotate head right
        body_action[0] = -1
    elif key == "n":  # tilt head down
        body_action[1] = 1
    elif key == "m":  # rotate head up
        body_action[1] = -1
```

Here, `base_action` is defined as `base_action = np.zeros([2])  # hardcoded for fetch robot`, meaning it doesn’t actually have a `base_action[2]`. To fix this, I modified the code as follows:

```python
if has_base:
    if key == "w":  # forward
        base_action[0] = 1
    elif key == "s":  # backward
        base_action[0] = -1
    elif key == "t":  # rotate counterclockwise
        base_action[1] = 1
    elif key == "e":  # rotate clockwise
        base_action[1] = -1
    elif key == "z":  # lift
        body_action[2] = 1
    elif key == "x":  # lower
        body_action[2] = -1
    elif key == "v":  # rotate head left
        body_action[0] = 1
    elif key == "b":  # rotate head right
        body_action[0] = -1
    elif key == "n":  # tilt head down
        body_action[1] = 1
    elif key == "m":  # rotate head up
        body_action[1] = -1
```

Can you confirm if this modification is correct? Also, why does pressing 'w' still affect other part of tobot?

---

StoneT2000:
@arth-shukla any idea? for the fetch robot

StoneT2000:
[...]

PR #996, opened 2025-04-09
[BugFix]: Fetch IK bug fix on GPU sim

arth-shukla:
Relevant to #958

Fixes fetch ee delta pose for GPU sim. Issues caused by dummy joints for hacked base makes IK fail to converge. Fixes:
- [x] use no-hacked-base URDF for the kinematic chain
- [x] fix gpu sim IK to ignore not-active-joints from Jacobian
- [x] directly call pinnochio for cpu ik, don't use sapiens func since this will not be maintained
- [x] allow using different link for root (e.g. if base is not root, like fetch)
- [x] fix fetch pd_ee_target_delta_pose ik


arth-shukla:
@StoneT2000 Re: CPU sim pinnochio model issue mentioned above

StoneT2000:
i dont remember why we use sapien's create pinnochio model function. We really should remove that dependency as sapien itself will not maintain it.

We should probably just create some function of articulations to do inverse kinematics in the first place maybe, or add it to the kinematics class, and call the actual pinnochio library?

arth-shukla:
@StoneT2000 CPU and GPU IK seem to work well for demo_random_control in PickCube on fetch and panda, and in ReplicaCAD on fetch. PyTest results are same as commit 74d70600f7ec8a8bfa46347d8fafa5725a12d6a9 (before ik fix commits). Are there other tests I should run?

mani_skill/agents/controllers/pd_ee_pose.py
```diff
@@ -60,7 +67,7 @@ def ee_pose(self):
 
     @property
     def ee_pose_at_base(self):
-        to_base = self.articulation.pose.inv()
+        to_base = self.root_link.pose.inv()
```
StoneT2000, line 70: why is this changed?

mani_skill/agents/controllers/utils/kinematics.py
```diff
@@ -79,14 +95,49 @@ def __init__(
...
         self.use_gpu_ik = False
-        # NOTE (stao): currently using the pinnochio that comes packaged with SAPIEN
-        self.qmask = torch.zeros(
-            self.articulation.max_dof, dtype=bool, device=self.device
+
+        with open(self.urdf_path, "r") as f:
+            xml = f.read()
+
+        joint_order = [
+            j.name  # f"joint_{j.child_link.index}"
```
StoneT2000: remove these comments

mani_skill/agents/controllers/utils/kinematics.py
```diff
@@ -79,14 +95,49 @@ def __init__(
...
+            # for j in sapien_articulation.active_joints
+        ]
+        link_order = [
+            l.name  # f"link_{l.index}"
+            for l in self.articulation.links
+            if l.name in self._kinematic_chain_link_names
+            # for l in sapien_articulation.links
+        ]
+
+        self.pmodel = PinocchioModel(xml, [0, 0, -9.81])
```
StoneT2000, line 113: can we avoid importing PinnochioModel? It is not installed by default actually

mani_skill/agents/controllers/utils/kinematics.py
```diff
@@ -180,7 +246,7 @@ def compute_ik(
             )
             if success:
                 return common.to_tensor(
-                    [result[self.active_ancestor_joint_idxs]], device=self.device
+                    [result[self.pmodel_controlled_joint_indices]], device=self.device
```
StoneT2000, line 247: so we have some joint qmask? And only use some IK solution dims?

mani_skill/agents/robots/fetch/fetch.py
```diff
@@ -27,6 +27,7 @@
 class Fetch(BaseAgent):
     uid = "fetch"
     urdf_path = f"{PACKAGE_ASSET_DIR}/robots/fetch/fetch.urdf"
+    urdf_stationary_path = f"{PACKAGE_ASSET_DIR}/robots/fetch/fetch_stationary.urdf"
```
StoneT2000: wouldn't his cause some problems? The base agent by default loads from urdf_path, this is only used for configuration purposes?

StoneT2000 (changes requested):

mani_skill/agents/controllers/pd_ee_pose.py
arth-shukla: Fetch's articulation root is always world origin bc of the extra links we add to model the base. So, I added an option to change the root link in the controller config.

mani_skill/agents/controllers/utils/kinematics.py
arth-shukla: Fixed

mani_skill/agents/controllers/utils/kinematics.py
arth-shukla: We can use PyTorch kinematics in cpu mode, but it's slower and less stable

mani_skill/agents/robots/fetch/fetch.py
arth-shukla: Yes `urdf_stationary_path` is just the fetch robot without the added joints for the mobile base. In the kinematics class I ignore joints in the ManiSkill articulation if they are not present in the urdf provided to the controller config. This assumes the ManiSkill articulation uses the same name for the joints, though

mani_skill/agents/controllers/utils/kinematics.py
arth-shukla: In this case:
- pmodel uses the controller config urdf (may be different than urdf used for articulation, as in fetch)
- `self.qmask` and `self.pmodel_controlled_joint_indices` are both the same (qmask is just a bool tensor, i.e. `self.qmask[self.pmodel_controlled_joint_indices] = 1`)
- `self.pmodel_controlled_joint_indices` is just `self.active_joint_indices` but reindexed to the controller_config_urdf

I replaced `self.active_ancestor_joint_idxs` since this seemed like a bug, it only works if `self.active_ancestor_joint_idxs == self.active_joint_indices`, since `self.set_drive_targets()` requires the drive targets to be the same order/length as `self.active_joint_indices`

mani_skill/agents/controllers/pd_ee_pose.py
LiuYiKai6: I think the root link should be the root of the urdf_stationary? Because in the your code, the Kinematics load model from urdf_stationary and the IK function input target_pose in the robot frame.

LiuYiKai6: And I have another question. The target_pose is in the robot frame (base_link) while the controlled joints only contain arm joints. And the torso_lift_joint between base_link and torso_lift_link ("arm base link") is controlled alone. It may affect the controller of the arm joints?

arth-shukla: @LiuYiKai6 thanks for pointing that out! while this shouldn't affect `pd_ee_delta_pose` (since it applies relative to current ee pose), this could affect `pd_ee_target_delta_pose`, since that keeps track of world pose transformation

can fix this by doing relative to torso link instead of base. will fix and test when I have some free time during the conf today

LiuYiKai6: why it shouldn't affect  pd_ee_delta_pose? the delta pose applies relative to current ee pose which is in the robot frame. what if the torso_lift_joint changes in the meantime? the step result of ee pose maybe not as expected.

arth-shukla: @LiuYiKai6 in testing pd_ee_delta_pose seems ok. since ik and torso pd for the current time step are computed at the same time, changes to the torso link qpos shouldn’t affect the IK for the delta pose

but it does break pd_ee_target_delta_pose, since this tracks target pose across time steps

@StoneT2000 this is fixed in the recent commit by using torso_lift_link as root for arm ik

arth-shukla:
@StoneT2000 double checking if this looks ok?

Jeffjewett27:
This issue has been affecting me. The commit has fixed my problems. The quality-of-life of the demo_manual_control_continuous.py is also much better. Thank you. I hope this can soon get merged.

StoneT2000 (approved):

PR #1022, opened 2025-04-25
Fixed PushCube policy not penalized for holding cube above target 

### LINKED ISSUE #1016: Push Cube Reward Does Not Penalize Z Axis When Success == False
Hi,  
When training RL agents on Push Cube, I noticed that sometimes, the agent can get stuck in a local minima where it will pick up the cube and hold it up above the target.  

**SAC State**   

https://github.com/user-attachments/assets/8e62a30c-f508-4382-8f60-9f11883c3036
   
**DrQ-V2 (RGB)**  

https://github.com/user-attachments/assets/b66c0883-7306-4074-b993-47e53c0516aa

  

When `success == True` the reward is set to the maximum value of 3 :  
```python
reward[info["success"]] = 3
```  
And the [success computation](https://github.com/haosulab/ManiSkill/blob/ba3562f5a6cfda7b7a534aac7aeb233d2c3cb362/mani_skill/envs/tasks/tabletop/push_cube.py#L183-L188) depends on the z-axis (good) :  
```
# success is achieved when the cube's xy position on the table is within the
# goal region's area (a circle centered at the goal region's xy position) and
# the cube is still on the surface
is_obj_placed = (
    torch.linalg.norm(
        self.obj.pose.p[..., :2] - self.goal_region.pose.p[..., :2], axis=1
    )
    < self.goal_radius
) & (self.obj.pose.p[..., 2] < self.cube_half_size + 5e-3)

return {
    "success": is_obj_placed,
}
```
But when `success == False` , the reward has the [following component](https://github.com/haosulab/ManiSkill/blob/ba3562f5a6cfda7b7a534aac7aeb233d2c3cb362/mani_skill/envs/tasks/tabletop/push_cube.py#L224-L227) :  
```python
obj_to_goal_dist = torch.linalg.norm(
    self.obj.pose.p[..., :2] - self.goal_region.pose.p[..., :2], axis=1
)
place_reward = 1 - torch.tanh(5 * obj_to_goal_dist)
reward += place_reward * reached
```  
Therefore the reward does not seem to use the z axis when `success == False` which could encourage the sub-optimal  behavior mentioned.  
   
# Suggestion
My suggestion would be to add the z axis to the reward for the case where `success == False` so that the agent is still incentivized to keep the cube on the table.

AlexandreBrown:
Fixes https://github.com/haosulab/ManiSkill/issues/1016

**The issue addressed**
The policy was not penalized for learning a sub-optimal policy that would learn to pick the cube and hold it above the target.    
This wouldn't always happen but could be replicated fairly easily and consistently.

**Note: This PR does not guarantee that a sub-optimal policy won't be learned, but it reduces the likelihood of learning a sub-optimal policy that learns to hold the cube above the target as the z axis is now penalized.**  

**Description of the change**
- Added z reward that encourages the cube to be on the table
- Similar to the previous xy reward, the z reward is only active when we first reached the push pose.
- The reward is gradually applied as the cube gets closer to the target pose.
  - This resulted in more consistent results than simply doing `z_reward * reached`
  - This prevents penalizing other optimal solutions ex: pick the cube then put it on the table, center of target after.
  - It still penalizes the sub-optimal behavior where the cube is held above the target center.  

# Results    
To confirm that this PR still results in good learning or even improved learning, I ran some experiments.  
## DrQ-V2 (RGB)     
Orange = Before this PR    
Blue = This PR   
Eval success at end, results is taken over 5 different seeds for each (before,after), range = std dev:  
image

Return curve (5 seeds again) :  
image

**Other result**
Blue = Before this PR, training diverges  
Green = After this PR, training no longer diverges and eval success rate remains higher throughout training
image
image    

Policy **at end of training** (different video = different seed) obtained with this PR (**not hand picking** the best, just looking for divergence or sub-optimal pattern) :  

https://github.com/user-attachments/assets/3384343d-9fa6-4c49-8100-bf85a51856fb
  

https://github.com/user-attachments/assets/72044635-20a8-4f09-954c-22092126d287

https://github.com/user-attachments/assets/4fef3ff7-c195-4266-960a-c83ec33b7865

We see that the policy shows the appropriate behavior.

## SAC State  
- Similar learning curves but final policy doesn't show sign of picking up the cube and holding it above the target.  
image  
  
## TD3 State  
- Same results as SAC State, learning curves still look good  
image

## DDPG State  
- Similar learning curves (slightly better but since this is only 1 seed, let's call it similar).    
image  

---  
# Alternatives Considered  
**In addition to the reward modification proposed in this PR**, I tested all the following **alternative** rewards formulation to include the z axis (across 4+ algos) :  
- Adding the z reward at all time (no masking + no gradual application of the reward) 
```python
def compute_dense_reward(self, obs: Any, action: Array, info: Dict):
        # Rest remains same

        desired_obj_z = self.cube_half_size
        current_obj_z = self.obj.pose.p[..., 2]
        z_deviation = torch.abs(current_obj_z - desired_obj_z)
        reward += 1 - torch.tanh(5 * z_deviation)

        # assign rewards to parallel environments that achieved success to the maximum of 3.
        reward[info["success"]] = 4
        return reward

    def compute_normalized_dense_reward(self, obs: Any, action: Array, info: Dict):
        # this should be equal to compute_dense_reward / max possible reward
        max_reward = 4.0
        return self.compute_dense_reward(obs=obs, action=action, info=info) / max_reward
```  
- Adding the z reward but using the `reached` mask to only apply it after we reached the push pose   
```python
def compute_dense_reward(self, obs: Any, action: Array, info: Dict):
        # ... Rest of code reamins the same  

        desired_obj_z = self.cube_half_size
        current_obj_z = self.obj.pose.p[..., 2]
        z_deviation = torch.abs(current_obj_z - desired_obj_z)
        z_reward = 1 - torch.tanh(5 * z_deviation)
        reward += z_reward * reached

        # assign rewards to parallel environments that achieved success to the maximum of 3.
        reward[info["success"]] = 4
        return reward

    def compute_normalized_dense_reward(self, obs: Any, action: Array, info: Dict):
        # this should be equal to compute_dense_reward / max possible reward
        max_reward = 4.0
        return self.compute_dense_reward(obs=obs, action=action, info=info) / max_reward
```   
- Adding the z reward + using the `reached` mask but applying a fixed weight to make the z reward less important than xy reward  
```python   
def compute_dense_reward(self, obs: Any, action: Array, info: Dict):
        # ...
        desired_obj_z = self.cube_half_size
        current_obj_z = self.obj.pose.p[..., 2]
        z_deviation = torch.abs(current_obj_z - desired_obj_z)
        z_reward = 1 - torch.tanh(5 * z_deviation)

        xy_weight = 1.0
        z_weight = 0.5

        reward += xy_weight * place_reward * reached
        reward += z_weight * z_reward * reached

        # assign rewards to parallel environments that achieved success to the maximum of 3.
        reward[info["success"]] = 4
        return reward

    def compute_normalized_dense_reward(self, obs: Any, action: Array, info: Dict):
        # this should be equal to compute_dense_reward / max possible reward
        max_reward = 4.0
        return self.compute_dense_reward(obs=obs, action=action, info=info) / max_reward
```
- Only using the negative L2 distance between the cube and the target pose (similar to what is done in https://robotics.farama.org/envs/fetch/push/)  
```python  
def compute_dense_reward(self, obs: Any, action: Array, info: Dict):
        obj_pose = self.obj.pose.p
        target_region_pose = self.goal_region.pose.p
        target_region_pose[..., 2] = self.cube_half_size
        obj_to_goal_dist = torch.linalg.norm(
            obj_pose - target_region_pose, axis=1
        )
        place_reward = 1 - torch.tanh(5 * obj_to_goal_dist)
        reward = place_reward

        # assign rewards to parallel environments that achieved success to the maximum of 3.
        reward[info["success"]] = 2
        return reward

    def compute_normalized_dense_reward(self, obs: Any, action: Array, info: Dict):
        # this should be equal to compute_dense_reward / max possible reward
        max_reward = 2.0
        return self.compute_dense_reward(obs=obs, action=action, info=info) / max_reward
```    
image  
image
  
Results show that this method makes learning slower and now state policies no longer converge.  
- Adding the z reward as part of the place reward (change xy reward to be xyz directly) but make the z index be between `half_cube_size` and the current cube's z position (not 0 because otherwise we'd be incentivizing the robot to push the cube through the table...).  
```python
 def compute_dense_reward(self, obs: Any, action: Array, info: Dict):
        tcp_push_pose = Pose.create_from_pq(
            p=self.obj.pose.p
            + torch.tensor([-self.cube_half_size - 0.005, 0, 0], device=self.device)
        )
        tcp_to_push_pose = tcp_push_pose.p - self.agent.tcp.pose.p
        tcp_to_push_pose_dist = torch.linalg.norm(tcp_to_push_pose, axis=1)
        reaching_reward = 1 - torch.tanh(5 * tcp_to_push_pose_dist)
        reward = reaching_reward
        
        reached = tcp_to_push_pose_dist < 0.01
        
        obj_pose = self.obj.pose.p
        target_region_pose = self.goal_region.pose.p.clone()
        target_region_pose[..., 2] = self.cube_half_size
        obj_to_goal_dist = torch.linalg.norm(
            obj_pose - target_region_pose, axis=1
        )
        place_reward = 1 - torch.tanh(5 * obj_to_goal_dist)
        reward += place_reward * reached

        # assign rewards to parallel environments that achieved success to the maximum of 3.
        reward[info["success"]] = 3
        return reward

    def compute_normalized_dense_reward(self, obs: Any, action: Array, info: Dict):
        # this should be equal to compute_dense_reward / max possible reward
        max_reward = 3.0
        return self.compute_dense_reward(obs=obs, action=action, info=info) / max_reward
```   
### Aggregated (Average over all algos for each method proposed)  
- Besides the negative L2 distance approach, all showed reasonable results but the selected on (separate z reward + reached mask + gradual application) resulted in the best curves and most consistent success rate results and judging by the videos it seems to help preventing the sub-optimal behavior although it might still occurr during training.  
**Original impl :**  
image
**This PR:**  
image

### DIFF
--- a/mani_skill/envs/tasks/tabletop/push_cube.py
+++ b/mani_skill/envs/tasks/tabletop/push_cube.py
@@ -45,7 +45,7 @@ class PushCubeEnv(BaseEnv):
     - the target goal region is marked by a red/white circular target. The position of the target is fixed to be the cube xy position + [0.1 + goal_radius, 0]
 
     **Success Conditions:**
-    - the cube's xy position is within goal_radius (default 0.1) of the target's xy position by euclidean distance.
+    - the cube's xy position is within goal_radius (default 0.1) of the target's xy position by euclidean distance and the cube is still on the table.
     """
 
     _sample_video_link = "https://github.com/haosulab/ManiSkill/raw/main/figures/environment_demos/PushCube-v1_rt.mp4"
@@ -226,12 +226,22 @@ class PushCubeEnv(BaseEnv):
         )
         place_reward = 1 - torch.tanh(5 * obj_to_goal_dist)
         reward += place_reward * reached
+        
+        # Compute a z reward to encourage the robot to keep the cube on the table
+        desired_obj_z = self.cube_half_size
+        current_obj_z = self.obj.pose.p[..., 2]
+        z_deviation = torch.abs(current_obj_z - desired_obj_z)
+        z_reward = 1 - torch.tanh(5 * z_deviation)
+        # We multiply the z reward by the place_reward and reached mask so that 
+        #   we only add the z reward if the robot has reached the desired push pose
+        #   and the z reward becomes more important as the robot gets closer to the goal.
+        reward += place_reward * z_reward * reached
 
         # assign rewards to parallel environments that achieved success to the maximum of 3.
-        reward[info["success"]] = 3
+        reward[info["success"]] = 4
         return reward
 
     def compute_normalized_dense_reward(self, obs: Any, action: Array, info: Dict):
         # this should be equal to compute_dense_reward / max possible reward
-        max_reward = 3.0
+        max_reward = 4.0
         return self.compute_dense_reward(obs=obs, action=action, info=info) / max_reward


AlexandreBrown:
Putting it as draft for now, I want to try a few more variations like using a gradual scale from 0 -> 5 until we reach the target boundary and then increase the z deviation scale aggressively to something like 20 to more aggressively penalize the z axis when above the target.  


UPDATE: No significant change observed, keeping the simpler z reward.

AlexandreBrown:
Might be worth trying to change the bonus for succes from reward = 4 to reward = 5 to make the incentive bigger.  
  
**UPDATE:** No significant change observed when using boost of reward = 5 on success instead of 4 :      
Results taken over 5 seeds, range = std dev.  
Orange = This PR  
Blue = With reward = 5 on success  
image  
image
    
Therefore I will stick to the original change and keep reward = 4 on success for simplicity.

AlexandreBrown:
Ready to review whenever you can @StoneT2000

PR #1038, opened 2025-05-03
[Feature] Add pointcloud ppo script

rrzhang139:
Added baseline ppo script for pointclouds. Useful for oracle (teacher-student) policies. 

Note: I wasn't able to fully run this script yet due to OOM errors, please let me know if there are any bugs.


StoneT2000:
One issue is for a true point cloud robot learning baseline we need to add code to support sampling configurations if obs_mode="pointcloud", or to use obs_mode="sensor_data" and parse into pointcloud data manually and then sample. Pointcloud data uses a ton of memory relative to the raw sensor data and typically we only sample some points.

rrzhang139:
Are there any libraries or code I can reference for adding this sampling configuration? I was thinking of something like passing them directly into the env.

`env = gym.make("PickCube-v1",
               obs_mode="pointcloud",
               point_cloud_sample=dict(method="fps", n_points=4096))`

rrzhang139:
@StoneT2000 Could you take a look at this new commit?

StoneT2000:
What are the results like for training with PCD on some tasks like PickCube or StackCube?

PR #1040, opened 2025-05-05
[Feature] optimize HDF5 loading in RGB DP baseline.

xucj98:
* Refactored the dataset to avoid preloading the entire dataset into memory during `__init__`.
* Deferred moving data to CUDA to avoid issues when `num_workers > 0` in DataLoader.
* Ensured each worker process opens its own HDF5 file to prevent shared file handle issues.
* Tuned HDF5 chunking settings to enable fast single-image access without loading entire trajectories.

⚠️ NOTE: This new version of the dataset requires HDF5 files generated with the updated chunking strategy.
Old-format files use chunking by trajectory (e.g. 24 frames per chunk), which makes single-frame access inefficient.
When using such files, reading one image will still trigger loading the entire chunk, causing significant slowdown.

### DIFF
--- a/examples/baselines/diffusion_policy/train_rgbd.py
+++ b/examples/baselines/diffusion_policy/train_rgbd.py
@@ -3,6 +3,7 @@ ALGO_NAME = "BC_Diffusion_rgbd_UNet"
 import os
 import random
 import time
+import h5py
 from collections import defaultdict
 from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


StoneT2000:
Since this is not compatible with previous datasets I cannot merge this now. But the idea is interesting, you are welcome to leave this as a post up in the discussion forums as well.

PR #1083, opened 2025-05-21
[BUG] Diffusion Policy Bug

rrzhang139:
Shifted the starting index of the action slice.

Ensures that the first predicted action (e.g. a_t) correctly follows the last observation (o_t), matching standard diffusion-policy convention (Chi et al.).


StoneT2000:
Have you tried running the training baselines yet?

PR #1086, opened 2025-05-22
Update doc to use correct vulkan package on ubuntu

akashsharma02:
Minor fix to install vulkan-tools instead of vulkan-utils.


StoneT2000 (approved):

StoneT2000:
Thanks!

PR #1089, opened 2025-05-26
Fixed cube_size wrong value

AlexandreBrown:
Removed `cube_size` since it is always divided by 2 and we already maintain `cube_half_size` and `cube_size` had the wrong value while `cube_half_size` is correct.

### DIFF
--- a/mani_skill/envs/tasks/tabletop/pull_cube_tool.py
+++ b/mani_skill/envs/tasks/tabletop/pull_cube_tool.py
@@ -44,7 +44,6 @@ class PullCubeToolEnv(BaseEnv):
     hook_length = 0.05
     width = 0.05
     height = 0.05
-    cube_size = 0.02
     arm_reach = 0.35
 
     def __init__(self, *args, robot_uids="panda", robot_init_qpos_noise=0.02, **kwargs):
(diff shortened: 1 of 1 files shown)


StoneT2000 (changes requested):
cube_half_size variable needs to be divided by 2 as well to maintain the same environment

AlexandreBrown:
@StoneT2000 Hey, the previous value was incorrect. The point of this PR is to fix that.  
Are you suggesting we should change the version number to `-v2` ?

StoneT2000:
I'd rather not have another task version, it becomes another thing to maintain and in research people get confused by task version numbers.

Since some research already uses this task it's better to leave the behavior the same (cube size same) even if it was initially a small typo.

AlexandreBrown:
Ok, closing this then, although I think improvements should still be possible for future research, I understand the desire to keep the repo easier to compare against.  
Maybe researchers should include the version of the libs they used @StoneT2000 , not something I see commonly reported in papers or even paper's repo sadly, some use loose dependency versions.

StoneT2000:
My current policy is for clear and obvious bugs (like very bad local minima in reward function or ill-defined success function or extremely inefficient sim due to some issue, e.g. your update for push cube earlier), we update the environment directly and hopefully not too much change is caused. I do recommend people to put the version number of the simulation in papers although it is not reported as you mention.

For smaller changes that don't affect research really (like changing the size of a cube in a task that can be solved) I will avoid making those updates.

Going forward we will probably be more stringent on the tasks included in this repo although there isn't a set policy yet, just recommendations.

Thanks again for all the contributions and discussion!

PR #1090, opened 2025-05-26
Add clamping to [-1, 1] for avoid nans in matrix_to_euler conversion

tylerlum:
I encountered strange nan issues when converting from quat_wxyz to euler_ZYX like below:

```
        pos = pose_w[:, :3]
        quat_wxyz = pose_w[:, 3:]
        matrix = quaternion_to_matrix(quat_wxyz)
        euler_ZYX = matrix_to_euler_angles(matrix, "ZYX")
        return torch.cat([pos, euler_ZYX], dim=-1)
```

I found that the issue was from numerical issues making matrix contain values greater than 1 (e.g., 1 + 1e-6). The domain of `torch.acos` and `torch.asin` are [-1, 1], and they return nan if given values outside of this range. Thus, this simply adds clamping to the inputs to these functions to make them robust to these numerical issues.

### DIFF
--- a/mani_skill/utils/geometry/rotation_conversions.py
+++ b/mani_skill/utils/geometry/rotation_conversions.py
@@ -291,10 +291,13 @@ def matrix_to_euler_angles(matrix: torch.Tensor, convention: str) -> torch.Tenso
     tait_bryan = i0 != i2
     if tait_bryan:
         central_angle = torch.asin(
-            matrix[..., i0, i2] * (-1.0 if i0 - i2 in [-1, 2] else 1.0)
+            torch.clamp(matrix[..., i0, i2], -1.0 + 1e-8, 1.0 - 1e-8)
+            * (-1.0 if i0 - i2 in [-1, 2] else 1.0)
         )
     else:
-        central_angle = torch.acos(matrix[..., i0, i0])
+        central_angle = torch.acos(
+            torch.clamp(matrix[..., i0, i0], -1.0 + 1e-8, 1.0 - 1e-8)
+        )
 
     o = (
         _angle_from_tan(


StoneT2000:
Wow if this is the case you should make a PR to the pytorch3d github repo as well. The code for rotation conversions is taken from there: https://github.com/facebookresearch/pytorch3d/blob/main/pytorch3d/transforms/rotation_conversions.py#L295-L310

That being said when are you observing this? Shouldn't this open happen if the matrix is ill defined in the first place?

tylerlum:
Thanks for the response!

Mathematically, there is no need for the clamping, as all rotation matrices should have values \in [-1, 1]. However, I believe numerical issues can result in matrix values being very slightly (~1e-6) out of range.

For example, here is the code for quaternion_to_matrix()

```
def quat_wxyz_to_matrix(quat_wxyz: torch.Tensor) -> torch.Tensor:
    """
    Convert rotations given as quaternions to rotation matrices.
    Args:
        quat_wxyz: quaternions with real part first,
            as tensor of shape (..., 4).
    Returns:
        Rotation matrices as tensor of shape (..., 3, 3).
    """
    r, i, j, k = torch.unbind(quat_wxyz, -1)
    two_s = 2.0 / (quat_wxyz * quat_wxyz).sum(-1)

    mat = torch.stack(
        (
            1 - two_s * (j * j + k * k),
            two_s * (i * j - k * r),
            two_s * (i * k + j * r),
            two_s * (i * j + k * r),
            1 - two_s * (i * i + k * k),
            two_s * (j * k - i * r),
            two_s * (i * k - j * r),
            two_s * (j * k + i * r),
            1 - two_s * (i * i + j * j),
        ),
        -1,
    )
    return mat.reshape(quat_wxyz.shape[:-1] + (3, 3))
```

I'm not an expert with exactly how that can happen, but if quat_wxyz is not PERFECTLY normalized, I believe this can happen.
For context, I had this error come up running with IsaacLab and getting the quaternions from their `body_quat_w` (copied the rotation utils from here). 

Is this clipping too expensive? If not, I think it's worth adding to avoid the numerical issues

StoneT2000:
Finally merged it. You are right, clipping is not expensive. I've also raised the question to pytorch3d repo but I somehow expect the answer will be they expect users to pass in matrices within range

tylerlum:
Thanks @StoneT2000 for the great work making things happen :)

PR #1091, opened 2025-05-27
[Bugfix] Vary sub-scene pick clutter object targets.

Gamenot:
Previous implementation selected target objects from only the last sub-scene (-1) causing scene cross-contamination if information from `target_object` was used. Original RL behavior after implementing the rewards generally resulted in reaching to a constant location on parallel environments due to referencing objects from a single scene.

Improvements:
- Use the batched RNG which scales already by `num_envs`.
- Generate the RNG for target selection only once for all sub-scenes.
- Select the target object for each sub-scene (i) from each sub-scene (i).


StoneT2000:
Thank you! I will try and review this early next week.

PR #1155, opened 2025-06-25
Fix partial reset support for digital twin tasks

zanghz21:
Previously, digital twin tasks did not support partial reset. The main issue was related to the computation of `info` during evaluation, which caused shape mismatches in partially reset environments.
```
File ".../mani_skill/envs/tasks/digital_twins/bridge_dataset_eval/put_on_in_scene.py", line 65, in evaluate
    info = super()._evaluate(
File ".../mani_skill/envs/tasks/digital_twins/bridge_dataset_eval/base_env.py", line 514, in _evaluate
    self.consecutive_grasp += is_src_obj_grasped
RuntimeError: The size of tensor a (3) must match the size of tensor b (20) at non-singleton dimension 0
```
Here, `3` corresponds to the number of partially reset environments, while `20` is the total number of environments.

This PR includes a quick fix to handle the shape mismatch and ensure compatibility with partial resets.


StoneT2000:
Thank you @zanghz21, code looks good

StoneT2000 (approved):

PR #1171, opened 2025-07-05
SO-101 Support

Viswesh-N:


mani_skill/assets/robots/so101/so101.urdf
```diff
@@ -337,11 +337,11 @@
 
   <!-- Joint from gripper to moving_jaw_so101_v1 -->
   <joint name="gripper" type="revolute">
-    <origin xyz="0.0202 0.0188 -0.0234" rpy="1.5708 -0.909 -1.41553e-15"/>
+    <origin xyz="0.0202 0.0188 -0.0234" rpy="1.5708 -5.24284e-08 -1.41553e-15"/>
```
StoneT2000: this e-08 value should just be 0?

mani_skill/assets/robots/so101/so101.urdf
```diff
@@ -337,11 +337,11 @@
...
     <parent link="gripper_link"/>
     <child link="moving_jaw_so101_v1_link"/>
     <axis xyz="0 0 1"/>
-    <limit effort="0" velocity="10" lower="-1.13778" upper="1.16081"/>
+    <limit effort="10" velocity="10" lower="-0.174533" upper="2.0944"/>
```
StoneT2000, line 344: we don't use effort but just checking these values of 10 are copied from the original so101 urdf, the new calibrated one here right https://github.com/TheRobotStudio/SO-ARM100/blob/main/Simulation/SO101/so101_new_calib.urdf

mani_skill/assets/robots/so101/so101.urdf
```diff
@@ -398,7 +398,7 @@
   <!-- Joint from upper_arm to lower_arm -->
   <!-- Note: 5-degree calibration offset applied to joint limits -->
   <joint name="elbow_flex" type="revolute">
-    <origin xyz="-0.11257 -0.028 1.73763e-16" rpy="-3.63608e-16 0.03 1.5708"/>
+    <origin xyz="-0.11257 -0.028 1.73763e-16" rpy="-3.63608e-16 8.74301e-16 1.4408"/>
```
StoneT2000: e-16 should just be 0

mani_skill/assets/robots/so101/so101.urdf
```diff
@@ -438,7 +438,7 @@
 
   <!-- Joint from base to shoulder -->
   <joint name="shoulder_pan" type="revolute">
-    <origin xyz="0.0388353 -8.97657e-09 0.0624" rpy="3.14159 4.18253e-17 -3.14159"/>
+    <origin xyz="0.0388353 -8.97657e-09 0.0624" rpy="3.14159 4.18253e-17 -3.20159"/>
```
StoneT2000: e-17 should be 0.

I noticed the official urdfs from the robotstudio used the e-17 values, but we should use 0 (and I'll tell them they should use 0 as well).

StoneT2000 (changes requested):

mani_skill/agents/robots/lerobot/manipulator.py
```diff
@@ -13,10 +13,10 @@
 from mani_skill.utils.structs.types import Array
 
 try:
-    from lerobot.common.cameras.camera import Camera
-    from lerobot.common.motors.motors_bus import MotorNormMode
-    from lerobot.common.robots.robot import Robot
-    from lerobot.common.utils.robot_utils import busy_wait
+    from cameras.camera import Camera
```
StoneT2000: this seems like a wrong import

mani_skill/agents/robots/lerobot/manipulator.py
```diff
@@ -58,6 +58,8 @@ def set_target_qpos(self, qpos: Array):
         # NOTE (stao): It seems the calibration from LeRobot has some offsets in some joints. We fix reading them here to match the expected behavior
         if self.real_robot.name == "so100_follower":
             qpos["elbow_flex.pos"] = qpos["elbow_flex.pos"] + 6.8
+        elif self.real_robot.name == "so101_follower":
```
StoneT2000: this is not needed the elif. In another PR we will also remove the SO100 one.

mani_skill/agents/robots/lerobot/manipulator.py
```diff
@@ -113,6 +115,8 @@ def get_qpos(self):
         # NOTE (stao): It seems the calibration from LeRobot has some offsets in some joints. We fix reading them here to match the expected behavior
         if self.real_robot.name == "so100_follower":
             qpos_deg["elbow_flex"] = qpos_deg["elbow_flex"] - 6.8
+        elif self.real_robot.name == "so101_follower":
```
StoneT2000: this elif is also not needed

mani_skill/assets/robots/so101/so101.srdf
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" ?>
+<robot name="so101">
+
+<!-- TODO: Check if this is correct -->
```
StoneT2000: can you verify this? Basically run the demo robot code in cpu sim, and open the contact tab and check there aren't any impossible self collisions being detected or considered

StoneT2000 (changes requested):

StoneT2000:
Is there a reason why the SO101 task is not set to be at the "normalized" 0 position? For consistency we want to reset to this "right angle, 0" position (even though in radians the joints wont be a multiple of np.pi/2). This way in the real world it also easier for users to do alignment as measuring multiples of 90 degrees is much easier.

We will do the same treatment for SO100 later as well once we fix that URDF rpy and limits as well

image

Screenshot from 2025-07-17 22-08-35

StoneT2000:
Another is we can decompose the mesh for the moving jaw a bit more on the SO101? To achieve the same quality as the SO100 one. You can even use the same collision mesh file probably.

Here is SO100 overlaid

image


Here is SO101 overlaid at the moment. Red circles mark some discrepancies

image

StoneT2000:
@Viswesh-N looking at the PR again. The collision meshes are still too off. For the moving jaw you may need two stl files to describe it and add both of them into the urdf, we load these both individually to form a non convex shape. SO100 modelling does this as well.

SRDF also needs updating. wrist_link and the moving_jaw_so101_v1_link are marked as potentially colliding when they can't actually collide.

StoneT2000:
image
finally this rest position seems off and not at the expected right angles. Can those be fixed as well?

StoneT2000:
One more thing, some of the values that should be 0 are really just those smaller than 1e-16 (at that precision it was probably a typo it was that small). Other values (like inertias which often are 1e-5 to 1e-8 range) should be kept the same.

StoneT2000:
closing for now

Issue #1168, opened 2025-07-04
Upload demo datasets for StackPyramid to HF.


StoneT2000:

chenyenru:
Added a pull request to the ManiSkill HF Dataset as well. Please take a look, thanks!

https://huggingface.co/datasets/haosulab/ManiSkill_Demonstrations/discussions/1/files

chenyenru:
Here's the result for DP state-based training on StackPyramid-v1: https://wandb.ai/yec01764/ManiSkill/runs/cu66scuf

PR #1172, opened 2025-07-05
[Feature] Add StackPyramid-v1 in motionplanning and il benchmark replay script

chenyenru:
Per https://github.com/haosulab/ManiSkill/issues/1168


StoneT2000:
Also forgot to mention this needs to be updated so users can download it with a reference to env id only:

https://github.com/haosulab/ManiSkill/blob/main/mani_skill/utils/download_demo.py#L27

@chenyenru

chenyenru:
thanks for the comprehensive check, just added it to the script!

StoneT2000 (approved):

PR #1212, opened 2025-08-01
fix: update root link retrieval

lvjonok:
Good day, thank you for this framework. I have been experimenting and found fairly easy bug that `get_state` method of base agent fails.

Would you like some test to be added for the BaseAgent or it is not necessary for this simple fix?

Best Regards,
Lev Kozlov

### DIFF
--- a/mani_skill/agents/base_agent.py
+++ b/mani_skill/agents/base_agent.py
@@ -358,7 +358,7 @@ class BaseAgent:
 
         # robot state
         root_link = self.robot.get_links()[0]
-        state["robot_root_pose"] = root_link.get_pose()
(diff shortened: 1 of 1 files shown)


StoneT2000:
Thanks for the fix! No test needed for this. 

Do you use this function and in what context? I figured most users use env.get_state_dict (and hence this bug being here for so long).

lvjonok:
> Thanks for the fix! No test needed for this.
> 
> Do you use this function and in what context? I figured most users use env.get_state_dict (and hence this bug being here for so long).

I am relatively new to work with ManiSkill, so when I needed to get qvel of the robot, the shortest path I found that time was through `agent.get_state()`.

Issue #1237, opened 2025-08-21
Motion Planning failing for SO100 arm


cmcgartoll:
Hi @StoneT2000!

Been looking to try out some SO100 training with Maniskill. Wanted to help contribute and setup the MotionPlanner + Teleop scripts for SO100 because it's helpful for me when checking out envs and simple interactions.

I was able to get it setup in my [fork](https://github.com/cmcgartoll/ManiSkill/tree/teleop-so100), and it works for some paths, but fails for a lot, especially ones involving rotation around the base motor. I tried motion planning with screw, RRTConnect, and RRTStar.

Here's an example of the working teleop:

https://github.com/user-attachments/assets/4ebc9c27-a944-4698-ac1c-7a377f175d71

Example error logs thrown by mplib for RRTConnect:
```
IK Failed! Distance 0.001396 is greater than threshold 0.001000.
Plan failed
IK Failed! Distance 0.002209 is greater than threshold 0.001000.
Plan failed
IK Failed! Distance 0.002745 is greater than threshold 0.001000.
Plan failed
IK Failed! Distance 0.002731 is greater than threshold 0.001000.
Plan failed
IK Failed! Distance 0.002099 is greater than threshold 0.001000.
Plan failed
IK Failed! Distance 0.001104 is greater than threshold 0.001000.
Plan failed
IK Failed! Distance 0.001031 is greater than threshold 0.001000.
Plan failed
IK Failed! Distance 0.001016 is greater than threshold 0.001000.
Plan failed
```

I was wondering if you've run into similar issues with mplib (or the SO100 in Maniskill) and if you know of any solutions to improve the motion planning! Thank you!

Jimntu:
Same issue, meet IK failed for so100 arm, looking forward to the answer!

chenyenru:
Hi! Could you print and visualize the new_pose here? https://github.com/haosulab/MPlib/blob/main/mplib/planner.py#L305C21-L305C25

It would help to compare both succeeding and failing cases (determined by whether min_dist<1e9) and see for the succeeding cases whether it was the quaternion that went really off (which will verify your hypothesis).

cmcgartoll:
Hi @chenyenru! Thanks for reaching out to help. Just ran a test and see some interesting behavior. 

[...]

PR #1247, opened 2025-08-28
[Feature]: Push Cube functionality for so100 and xarm6

cmcgartoll:
This PR is built from recommendation from @chenyenru in issue https://github.com/haosulab/ManiSkill/issues/1237 to work on motionplanning solutions for so100.

PR contains following:
- Updated mani_skill/envs/tasks/tabletop/push_cube.py to handle multiple arm configs
- Added PushCube mp solution for so100 [Example video](https://github.com/user-attachments/assets/7e36c338-210e-43af-9ac4-b9ad84db10b9)
- Tweaked PushCube mp solution for xarm6 to prevent it from hitting table [Example video](https://github.com/user-attachments/assets/a93174cd-f3e4-4b34-8a50-317217137363)

### DIFF
--- a/mani_skill/envs/tasks/tabletop/push_cube.py
+++ b/mani_skill/envs/tasks/tabletop/push_cube.py
@@ -73,6 +73,7 @@ class PushCubeEnv(BaseEnv):
         else:
             cfg = PUSH_CUBE_CONFIGS["panda"]
         self.goal_radius = cfg["goal_radius"]
+        self.goal_distance = cfg["goal_distance"]
         self.cube_half_size = cfg["cube_half_size"]
         self.cube_spawn_center = cfg["cube_spawn_center"]
         self.cube_spawn_half_size = cfg["cube_spawn_half_size"]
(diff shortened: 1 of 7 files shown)


StoneT2000:
Thank you for the efforts!

Unfortunately I'm not sure if I can accept the PR. PushCube is meant to be kept as simple as possible since it's also used as a tutorial environment to guide users on how to make custom environments.

PickCube is kept more complex since that environment is also used to test the manipulation stability of a manipulation robot.

cmcgartoll:
Thanks! Yeah makes sense since it is the example script. 

Just a thought, could we maybe change the current example push_cube.py script into a push_cube_example.py script (since it's the only example), and then have a standard push_cube.py script without all the comments + with multi-robot config capabilities? 

Or, alternatively register my changes as a PushCube-v2 env and leave the current v1 setup as is? 

It'd be nice to have that ability to use SO100/many other robots in the push_cube env. Just some thoughts and would be interested to hear what you think, thanks!

StoneT2000:
I think a config system warrants some necessity to keep more consistency between environments. A well designed config system is difficult, the one used for pick cube is really very specific for testing and not a good design.

ManiSkill doesn't intend to force a particular configuration system that you have to use to modify an environment. Everyone has their own preferences on what works best for them. For now if you really want to use such a config system I recommend you make these environments in your own repository.

If you do make a nice config system interface for working with all the maniskill envs that involves less python code and more config files (similar to isaac lab manager system) that would be pretty cool! Happy to link to your repo in that case in our docs.

cmcgartoll:
I'll check out Isaac Lab's manager system, and explore an improved config system. Thanks!

PR #1255, opened 2025-09-04
[Docs] Tasks mistakenly documented as supporting MotionPlanning

dfloreaa:
### Description  
Fix an inconsistency in [the documentation](https://maniskill.readthedocs.io/en/latest/tasks/table_top_gripper/index.html) where tasks were featured as having motion planning support, even though they are not listed in the [motionplanning task dictionary](https://github.com/haosulab/ManiSkill/blob/main/mani_skill/examples/motionplanning/panda/run.py).


StoneT2000:
I don't quite follow why this is needed. The docs add a checkmark there if the task has demonstrations, not a motion planning solution. Some demos are generated via RL.

dfloreaa:
You're completely right. My bad, I was misled by the terminology, do you think adding a new row for motion planning may be useful? It's not really documented anywhere but in the file itself (and the error message for when you try to actually generate the trajectories)

StoneT2000:
We can consider it. Something that would be useful is for each task to have a checkmark for whether 
1. An RL baseline exists for it (and can kind of work)
2. motion planning baseline exists

UI wise I am not sure how to organize this. We could also include these details on some page summarizing baselines

PR #1293, opened 2025-10-09
Adding XLeRobot as officially supported robot

Vector-Wangel:


StoneT2000:
Awesome work!

One issue left is just this:
image

there to be some impossible contacts (doesn't collide but it is slowing the sim down) shown above. can you disable those in the SRDF?

Vector-Wangel:
> Awesome work!
> 
> One issue left is just this: image
> 
> there to be some impossible contacts (doesn't collide but it is slowing the sim down) shown above. can you disable those in the SRDF?

Thanks! Just updated the asset link to disable the contacts between base link and "Rotation_pitch (both 1 & 2)".

StoneT2000 (approved):

StoneT2000:
collisions could possibly be optimized a bit more but should not affect results in general. Thanks, now merged into maniskill nightly!

PR #1297, opened 2025-10-18
[Feature] Add new dextrous hand tasks, more floating dextrous hand examples, support optionally saving during env resets for record episode wrapper

liuyulinn:
- Add a dexterity environment and a floating robotic hand. 
- Update the recorder and trajectory replay modules to support teleoperation functions.
- Code clean up.

### DIFF
--- a/mani_skill/agents/robots/floating_ability_hand/floating_ability_hand.py
+++ b/mani_skill/agents/robots/floating_ability_hand/floating_ability_hand.py
@@ -1,0 +1,217 @@
+from copy import deepcopy
+from typing import Dict, Tuple
+
+import numpy as np
+import sapien
+import sapien.physx as physx
+
+from mani_skill import PACKAGE_ASSET_DIR
+from mani_skill.agents.base_agent import BaseAgent, Keyframe
+from mani_skill.agents.controllers import *
+from mani_skill.agents.registration import register_agent
+from mani_skill.utils import sapien_utils
+
+
+@register_agent()
+class FloatingAbilityHandRight(BaseAgent):
+    uid = "floating_ability_hand_right"
+    urdf_path = f"{PACKAGE_ASSET_DIR}/robots/ability_hand/ability_hand_right_floating.urdf"
+    urdf_config = dict(
+        _materials=dict(
+            front_finger=dict(
+                static_friction=2.0, dynamic_friction=1.5, restitution=0.0
+            )
+        ),
+        link=dict(
+            thumnb_L2=dict(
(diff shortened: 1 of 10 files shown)


mani_skill/agents/robots/floating_ability_hand/floating_ability_hand.py
```diff
@@ -0,0 +1,217 @@
...
+
+from mani_skill import PACKAGE_ASSET_DIR
+from mani_skill.agents.base_agent import BaseAgent, Keyframe
+from mani_skill.agents.controllers import *
+from mani_skill.agents.registration import register_agent
+from mani_skill.utils import sapien_utils
+
+
+@register_agent()
+class FloatingAbilityHandRight(BaseAgent):
```
StoneT2000, line 16: from running the code it seems there is a missing mesh (one of the fingers), we need to update the ability hand asset files

the URDF is ill-defined there actually. Mesh exists, but path is wrong

mani_skill/trajectory/replay_trajectory.py
```diff
@@ -399,6 +399,7 @@ def _main(
     # Load associated json
     json_path = traj_path.replace(".h5", ".json")
     json_data = io_utils.load_json(json_path)
+    env_kwargs["sim_config"]["scene_config"]["gravity"] = np.array(env_kwargs["sim_config"]["scene_config"]["gravity"])
```
StoneT2000, line 402: the correct fix is to add a scene config parser somewhere in the BaseEnv class. It seems

```
        self.sim_config = dacite.from_dict(data_class=SimConfig, data=merged_gpu_sim_config, config=dacite.Config(strict=True))

``` 
isn't doing the job? I'll let this pass for now and add an issue to fix this later

mani_skill/envs/tasks/dexterity/insert_flower.py
```diff
@@ -0,0 +1,189 @@
+from pathlib import Path
+from typing import Any, Dict, List, Union
+
+import numpy as np
+import sapien
+import sapien.physx as physx
+import torch
+import torch.nn.functional as F
```
StoneT2000: remove the unused imports (F, randomization)

StoneT2000 (changes requested):

mani_skill/envs/tasks/dexterity/insert_flower.py
StoneT2000: actually there's more than that, there's a bunch of other ones too.

StoneT2000 (approved):

PR #1392, opened 2026-03-08
Fix potential XXE and environment signature mismatch

RinZ27:
This update addresses a potential security risk in the MJCF loader and fixes a runtime inconsistency in the environment templates.

Switching to `defusedxml.ElementTree` in `_mjcf_loader.py` ensures that MJCF XML parsing is protected against XML External Entity (XXE) attacks. This is particularly relevant for researchers loading model assets from community-provided sources.

Additionally, the `evaluate` and `set_state_dict` methods in `mani_skill/envs/template.py` have been updated to match the signature of `BaseEnv`. These changes prevent `TypeError` exceptions during execution in vectorized simulation environments where `env_idx` is required or additional arguments are not expected. Verified these fixes locally to ensure parity with existing core logic.

### DIFF
--- a/mani_skill/envs/template.py
+++ b/mani_skill/envs/template.py
@@ -155,7 +155,7 @@ class CustomEnv(BaseEnv):
     the code below all impact some part of `self.step` function
     """
 
-    def evaluate(self, obs: Any):
+    def evaluate(self):
         # this function is used primarily to determine success and failure of a task, both of which are optional. If a dictionary is returned
         # containing "success": bool array indicating if the env is in success state or not, that is used as the terminated variable returned by
         # self.step. Likewise if it contains "fail": bool array indicating the opposite (failure state or not) the same occurs. If both are given
@@ -198,8 +198,8 @@ class CustomEnv(BaseEnv):
         # state["goal_pos"] = add_your_non_sim_state_data_here
         return state
 
-    def set_state_dict(self, state):
+    def set_state_dict(self, state, env_idx: torch.Tensor = None):
         # this function complements get_state and sets any non simulation state related data correctly so the environment behaves
         # the exact same in terms of output rewards, observations, success etc. should you reset state to a given state and take the same actions
-        self.goal_pos = state["goal_pos"]
-        super().set_state_dict(state)
+        # self.goal_pos = state["goal_pos"]
+        super().set_state_dict(state, env_idx)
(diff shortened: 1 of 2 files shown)


mani_skill/envs/template.py
```diff
@@ -198,8 +198,8 @@ def get_state_dict(self):
         # state["goal_pos"] = add_your_non_sim_state_data_here
         return state
 
-    def set_state_dict(self, state):
+    def set_state_dict(self, state, env_idx: torch.Tensor = None):
         # this function complements get_state and sets any non simulation state related data correctly so the environment behaves
         # the exact same in terms of output rewards, observations, success etc. should you reset state to a given state and take the same actions
-        self.goal_pos = state["goal_pos"]
-        super().set_state_dict(state)
+        # self.goal_pos = state["goal_pos"]
```
StoneT2000, line 204: thanks for updating the template as well here. Can this line not be uncommented? It's just to show an example of how one could modify the state dict

mani_skill/utils/building/_mjcf_loader.py
```diff
@@ -39,14 +39,14 @@
...
 import re
-import xml.etree.ElementTree as ET
 from collections import defaultdict
 from copy import deepcopy
 from dataclasses import dataclass
 from functools import reduce
 from typing import Any, Literal, Tuple, Union
 from xml.etree.ElementTree import Element
 
+import defusedxml.ElementTree as ET
```
StoneT2000, line 49: is this package the standard today? It seems it is no longer maintained

StoneT2000 (changes requested):

mani_skill/envs/template.py
RinZ27: I've commented it out precisely for that reason—having it uncommented in the template would cause a `KeyError` if anyone copied the code without defining `goal_pos` first. Keeping it as a comment still serves as a perfect example while making the template "runnable" by default.

mani_skill/utils/building/_mjcf_loader.py
RinZ27: While it hasn't had frequent updates recently, `defusedxml` is actually considered a "stable/mature" project and is maintained by Christian Heimes (a Python core developer). It's still the officially recommended solution in the Python documentation for protecting against XXE in `ElementTree`. Given it's a drop-in replacement, it's the safest way to harden this loader without complex refactoring. @StoneT2000

PR #1397, opened 2026-03-17
Add camera pose editors plugins

fracapuano:
Adds an interactive camera-pose editor directly in the renderer, so that users can modify (and read!) the pose of different cameras more easily. Demo 👇 

Run the same command with:
```bash
 python3 -m mani_skill.examples.demo_random_action --render-mode human --pause --edit-camera-poses
```

As an aside: wow, Codex (@vaibhavs10, 🥹)---simply wow.

### DIFF
--- a/mani_skill/envs/sapien_env.py
+++ b/mani_skill/envs/sapien_env.py
@@ -1344,6 +1344,13 @@ class BaseEnv(gym.Env):
     def _ensure_viewer_camera_control_plugin(self):
         if self._viewer is None:
             return
+        existing_plugin = sapien_utils.get_obj_by_type(
+            self._viewer.plugins, ViewerCameraControlPlugin, is_unique=False
+        )
+        if isinstance(existing_plugin, list):
+            self._viewer_camera_control_plugin = existing_plugin[0]
+        elif existing_plugin is not None:
+            self._viewer_camera_control_plugin = existing_plugin
         if self._viewer_camera_control_plugin is None:
             self._viewer_camera_control_plugin = ViewerCameraControlPlugin(self)
             self._viewer.plugins.append(self._viewer_camera_control_plugin)
--- a/mani_skill/envs/sapien_env.py
+++ b/mani_skill/envs/sapien_env.py
@@ -39,6 +39,9 @@ from mani_skill.utils import common, gym_utils, sapien_utils, tree
 from mani_skill.utils.structs import Actor, Articulation
 from mani_skill.utils.structs.pose import Pose
 from mani_skill.utils.structs.types import Array, SimConfig
+from mani_skill.utils.visualization.viewer_camera_control import (
+    ViewerCameraControlPlugin,
+)
 from mani_skill.utils.visualization.misc import tile_images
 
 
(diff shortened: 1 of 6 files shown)


mani_skill/utils/visualization/viewer_camera_control.py
```diff
@@ -0,0 +1,411 @@
...
+
+from mani_skill.sensors.camera import Camera
+from mani_skill.utils.structs.pose import Pose
+
+if TYPE_CHECKING:
+    from mani_skill.envs.sapien_env import BaseEnv
+
+
+_VIEWER_GL_POSE = sapien.Pose([0, 0, 0], [-0.5, -0.5, 0.5, 0.5])
+_CAMERA_LINESET_VERTICES = [
```
StoneT2000, line 20: Can the _CAMERA_LINESET_VERTICES variable be made a single line (and add a comment to ignore lint checks for that line)

mani_skill/utils/visualization/viewer_camera_control.py
```diff
@@ -0,0 +1,411 @@
+from __future__ import annotations
+
```
StoneT2000, line 2: nice use of the sapien plugin class.

I think this won't be the only non sapien provided plugin we may wish to provide given there could be some other useful features (e.g. reward tracking à la mujoco/mjx tooling).

Moreover we kind of shoehorned some other viewer related code into this folder https://github.com/haosulab/ManiSkill/tree/main/mani_skill/render which I never really liked.

Thus, I think we should make a new folder mani_skill/viewer and then another folder mani_skill/viewer/plugins and add this plugin here.

mani_skill/examples/demo_random_action.py
```diff
@@ -49,6 +49,9 @@ class Args:
     pause: Annotated[bool, tyro.conf.arg(aliases=["-p"])] = False
     """If using human render mode, auto pauses the simulation upon loading"""
 
+    edit_camera_poses: bool = False
```
StoneT2000, line 52: The demo random action file should not be demoing specific plugins like camera editing. I recommend making a new function to demonstrate how to add plugins to a viewer by a user in code `examples/demo_plugins.py`. Let user specify in the CLI which plugins to load, and print a message to the user to tell them to read the demo's code for how to use non-default plugins

We can include some widely used plugins in the future as part of the default set of plugins added to sapien in maniskill (and this could be one of them, at the moment I'd hold off from making it a default).

mani_skill/envs/sapien_env.py
```diff
@@ -115,6 +118,9 @@ class BaseEnv(gym.Env):
         enhanced_determinism (bool): By default this is False and env resets will reset the episode RNG only when a seed / seed list is given.
             If True, the environment will reset the episode RNG upon each reset regardless of whether a seed is provided.
             Generally enhanced_determinisim is not needed and users are recommended to pass seeds into the env reset function instead.
+
+        enable_camera_pose_editing (bool): If True, the human viewer exposes an in-window camera editor that lets you click camera frustums
```
StoneT2000, line 122: given I think we should make new plugins a bit more standardized, we should not add specific viewer plugin related code here like one just for camera editing.

Instead, we have two (not mutually exclusive) options
- Add a function `attach_viewer_plugin` that accepts a created plugin object or maybe the plugin class and adds it to the viewer once viewer is created (we will need to maintain a list of plugins).
- In the BaseEnv init args, add a argument `plugins` and you can pass in a list of plugins that we add upon viewer creation

StoneT2000 (changes requested):
Incredible work! This has been a requested feature in the past by some people, glad you made a great implementation for it! See my comments

fracapuano:
(sorry for the gymnasium confusion, pushed to the wrong branch!)

PR #1430, opened 2026-06-07
fix: normalize episode_seed type to prevent TypeError in replay_trajectory

### LINKED ISSUE #1413: Adding observations and rewards to trajectories gives an error in Maniskill 3
```
python -m mani_skill.trajectory.replay_trajectory \
  --traj-path path/to/trajectory.h5 \
  --record-rewards --reward-mode="normalized_dense" -o rgb \
  --use-env-states \
  --save-traj
```

I wanted to use the Maniskill `replay_trajectory` library to store the rewards and observations associated with trajectories stored in a `.h5` file. However, the above command gave me the following error:

```
(base) root@gayathri-manip-pod-1-2-http:~/.maniskill/demos/PickCube-v1/motionplanning# python -m mani_skill.trajectory.replay_trajectory --traj-path /root/.maniskill/demos/PickCube-v1/motionplanning/trajectory.h5 --record-rewards --reward-mode="normalized_dense" -o rgb --use-env-states --save-traj
0step [00:00, ?step/s]WARNING:root:shader_dir argument will be deprecated after ManiSkill v3.0.0 official release. Please use sensor_configs/human_render_camera_configs to set shaders.
0step [00:02, ?step/s, control_mode=pd_joint_pos, obs_mode=rgb]/opt/conda/lib/python3.9/site-packages/gymnasium/core.py:311: UserWarning: WARN: env.obs_mode to get variables from other wrappers is deprecated and will be removed in v1.0, to get this variable you can do env.unwrapped.obs_mode for environment variables or env.get_wrapper_attr('obs_mode') that will search the reminding wrappers.
logger.warn(
/opt/conda/lib/python3.9/site-packages/gymnasium/core.py:311: UserWarning: WARN: env.control_mode to get variables from other wrappers is deprecated and will be removed in v1.0, to get this variable you can do env.unwrapped.control_mode for environment variables or env.get_wrapper_attr('control_mode') that will search the reminding wrappers.
logger.warn(
/opt/conda/lib/python3.9/site-packages/gymnasium/core.py:311: UserWarning: WARN: env.device to get variables from other wrappers is deprecated and will be removed in v1.0, to get this variable you can do env.unwrapped.device for environment variables or env.get_wrapper_attr('device') that will search the reminding wrappers.
logger.warn(
/opt/conda/lib/python3.9/site-packages/gymnasium/core.py:311: UserWarning: WARN: env.max_episode_steps to get variables from other wrappers is deprecated and will be removed in v1.0, to get this variable you can do env.unwrapped.max_episode_steps for environment variables or env.get_wrapper_attr('max_episode_steps') that will search the reminding wrappers.
logger.warn(
Replaying traj_0: : 0step [00:03, ?step/s, control_mode=pd_joint_pos, obs_mode=rgb]Traceback (most recent call last):
File "/opt/conda/lib/python3.9/runpy.py", line 197, in _run_module_as_main
return _run_code(code, main_globals, None,
File "/opt/conda/lib/python3.9/runpy.py", line 87, in _run_code
exec(code, run_globals)
File "/workspace/ManiSkill/mani_skill/trajectory/replay_trajectory.py", line 335, in <module>
main(parse_args())
File "/workspace/ManiSkill/mani_skill/trajectory/replay_trajectory.py", line 329, in main
_main(args)
File "/workspace/ManiSkill/mani_skill/trajectory/replay_trajectory.py", line 189, in _main
assert reset_kwargs["seed"] == ep["episode_seed"][0]
TypeError: 'int' object is not subscriptable
Replaying traj_0: : 0step [00:03, ?step/s, control_mode=pd_joint_pos, obs_mode=rgb]
```

I did the following to fix it:

In `trajectory/replay_trajectory.py`:
```
episode_seed = ep["episode_seed"]
if isinstance(episode_seed, np.ndarray):
        episode_seed = episode_seed.item() if episode_seed.shape == () else episode_seed[0]
elif isinstance(episode_seed, (list, tuple)):
        episode_seed = episode_seed[0]
if "seed" in reset_kwargs:
        assert reset_kwargs["seed"] == episode_seed
else:
       reset_kwargs["seed"] = episode_seed
```

Ottohere-Mourn:
## Summary

- Fixed `TypeError: 'int' object is not subscriptable` when using `--use-env-states` with `--record-rewards` and observation recording flags in `replay_trajectory`
- The `episode_seed` field can be stored as a numpy array, list, or scalar int depending on how the trajectory was collected
- Added normalization logic to convert `episode_seed` to a scalar int before comparison or assignment
- Handles numpy arrays (0-d and 1-d), lists, tuples, and plain ints

Fixes #1413

## Test Plan

- [ ] `python -m mani_skill.trajectory.replay_trajectory --traj-path <path> --record-rewards --use-env-states --save-traj` no longer raises TypeError
- [ ] Existing replay functionality continues to work without `--use-env-states`

### DIFF
--- a/mani_skill/trajectory/replay_trajectory.py
+++ b/mani_skill/trajectory/replay_trajectory.py
@@ -94,6 +94,14 @@ class ReplayResult:
 
 def sanity_check_and_format_seed(episode):
     """sanity checks the trajectory seed aligns with the episode seed. reformats the reset kwargs seed if missing or formatted wrong"""
+    # normalize episode_seed to a scalar int (may be stored as numpy array or list)
+    episode_seed = episode["episode_seed"]
+    if isinstance(episode_seed, np.ndarray):
+        episode_seed = episode_seed.item() if episode_seed.ndim == 0 else episode_seed[0]
+    elif isinstance(episode_seed, (list, tuple)):
+        episode_seed = episode_seed[0]
+    episode["episode_seed"] = int(episode_seed)
+
     if "seed" in episode["reset_kwargs"]:
         if isinstance(episode["reset_kwargs"]["seed"], list):
 
--- a/mani_skill/trajectory/replay_trajectory.py
+++ b/mani_skill/trajectory/replay_trajectory.py
@@ -97,8 +97,19 @@ def sanity_check_and_format_seed(episode):
     # normalize episode_seed to a scalar int (may be stored as numpy array or list)
     episode_seed = episode["episode_seed"]
     if isinstance(episode_seed, np.ndarray):
-        episode_seed = episode_seed.item() if episode_seed.ndim == 0 else episode_seed[0]
+        if episode_seed.ndim == 0:
+            episode_seed = episode_seed.item()
+        else:
+            assert episode_seed.size == 1, (
+                f"found multiple seeds for one trajectory (id={episode['episode_id']}) "
+                "in episode_seed which means it is ambiguous which seed to use"
+            )
+            episode_seed = episode_seed.reshape(-1)[0]
     elif isinstance(episode_seed, (list, tuple)):
+        assert len(episode_seed) == 1, (
+            f"found multiple seeds for one trajectory (id={episode['episode_id']}) "
+            "in episode_seed which means it is ambiguous which seed to use"
+        )
         episode_seed = episode_seed[0]
     episode["episode_seed"] = int(episode_seed)
 


ErenAta16 (commented):
The crash is real and the normalization handles every shape the description lists. I ran the twelve input shapes I could think of through it:

```
scalar int      -> 42        0-d ndarray  -> 42        1-d ndarray [42] -> 42
list [42]       -> 42        tuple (42,)  -> 42        np.int64(42)     -> 42
```

All correct, and the single-element unwrapping is exactly what was missing.

## The multi-element branch contradicts the policy four lines below it

The rest of `sanity_check_and_format_seed` already has an opinion about a seed field holding more than one value:

```python
if isinstance(episode["reset_kwargs"]["seed"], list):
    assert (
        len(episode["reset_kwargs"]["seed"]) == 1
    ), f"found multiple seeds for one trajectory (id={episode['episode_id']}) in the reset kwargs which means it is ambiguous which seed to use"
```

Refuse, and say why. This PR takes the opposite line on the sibling field:

```python
episode_seed = episode_seed.item() if episode_seed.ndim == 0 else episode_seed[0]
```

Silently keep the first, discard the rest, and write the scalar back into `episode`. Two fields, the same ambiguity, opposite policies, in the same function.

## What that costs, measured

Running the whole function on `main` and with this patch:

| `episode_seed` | `reset_kwargs["seed"]` | `main` | this PR |
|---|---|---|---|
| `[42, 99]` | `42` | `ValueError: truth value of an array with more than one element is ambiguous` | **passes**, `episode_seed=42` |
| `[42, 99]` | `99` | same `ValueError` | `AssertionError: found mismatch...` |
| `[42, 99]` | absent | propagates `[42 99]` | propagates `42` |

Row one is the one I'd want changed. The assert two lines down exists specifically to catch a trajectory whose recorded seed disagrees with its reset kwargs. With a multi-seed `episode_seed`, that check now compares against the first element only, so a trajectory carrying two seeds passes a check written to reject exactly that ambiguity. On `main` it fails loudly; noisy, but it fails.

Row three is quieter and possibly worse: with no `reset_kwargs["seed"]`, the function copies `episode_seed` across. `main` copies the whole array, this copies `42`, so the replay resets with a different seed than was recorded, with nothing printed.

Mirroring the existing policy keeps the fix and closes this:

```python
if isinstance(episode_seed, np.ndarray):
    if episode_seed.ndim == 0:
        episode_seed = episode_seed.item()
    else:
        assert episode_seed.size == 1, (
            f"found multiple seeds for one trajectory (id={episode['episode_id']}) "
            "in episode_seed which means it is ambiguous which seed to use"
        )
        episode_seed = episode_seed.reshape(-1)[0]
```

Same shape as the message already in the function, so a user hitting either gets the same explanation.

## Three inputs still raise, just differently

```
empty ndarray  -> IndexError: index 0 is out of bounds for axis 0 with size 0
empty list     -> IndexError: list index out of range
2-d ndarray    -> TypeError: only length-1 arrays can be converted to Python scalars
None           -> TypeError: int() argument must be a string, a bytes-like...
```

None of these are worse than `main`, and an empty seed is nonsense input either way. Worth knowing they exist rather than assuming the normalization is total — if the intent is that this function never raises a raw exception on a malformed trajectory, `size == 1` above would catch the empty and 2-d cases too and give them the same named error.

Also: since this mutates `episode` in place and the mutation is what everything downstream sees, it's worth a word in the docstring. The function already reformats `reset_kwargs`, so it isn't a new kind of side effect, but `episode_seed` changing type from array to int is easy to miss from the call site.

Ottohere-Mourn (commented):
Thanks for the thorough review! I've addressed the feedback:

- Multi-element ndarray seeds now raise an  with the same message format as the existing  seed check, instead of silently discarding extra elements
- Multi-element list/tuple seeds similarly raise on ambiguous input
- The 0-d ndarray scalar path is unchanged

The  normalization now follows the same policy as the  seed handling a few lines below.

Ottohere-Mourn (commented):
Thanks for the thorough review! I've addressed the feedback:

- Multi-element ndarray seeds now raise an AssertionError with the same message format as the existing reset_kwargs seed check, instead of silently discarding extra elements
- Multi-element list/tuple seeds similarly raise on ambiguous input
- The 0-d ndarray scalar path is unchanged

The episode_seed normalization now follows the same policy as the reset_kwargs seed handling a few lines below.

PR #1449, opened 2026-06-10
[BugFix] Explicitly register viewer plugins so GUI panels appear

Kaweees:
`create_viewer()` in `mani_skill/utils/sapien_utils.py` constructs the SAPIEN `Viewer` without passing a `plugins` argument. With recent SAPIEN versions, an unspecified `plugins` list no longer falls back to the full set of UI windows, so the interactive viewer comes up missing its control panels. This PR explicitly passes the standard plugin windows to restore the full visualization UI.

### DIFF
--- a/mani_skill/utils/sapien_utils.py
+++ b/mani_skill/utils/sapien_utils.py
@@ -12,6 +12,14 @@ import sapien.physx as physx
 import sapien.render
 import torch
 from sapien.utils import Viewer
+from sapien.utils.viewer.contact_window import ContactWindow
+from sapien.utils.viewer.control_window import ControlWindow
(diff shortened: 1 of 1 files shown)


StoneT2000:
@Kaweees what version of SAPIEN are you using?

Their latest code seems to still install all the windows: https://github.com/haosulab/SAPIEN/blob/41921b2243538f9a4a6b3d0e7199a5e9f6688e02/python/py_package/utils/viewer/viewer.py#L52-L62

Kaweees:
I was using a version prior to SAPIEN 3.0.3, which installs all the windows as you said. Closing. Could you check my messages on Discord to you regarding a pull request I created for aarch64 support?

PR #1457, opened 2026-06-16
Fix scalar-seeded main RNG expansion shrinking the episode seed batch

### LINKED ISSUE #1456: Scalar seed expansion gives wrong RNG batch size for vectorized envs
There seems to be a bug in vectorized environments when a scalar seeded reset is followed by an unseeded reset with reconfiguration enabled.

The seeded reset initializes the episode seeds with the correct length, but the following unseeded reset changes `_episode_seed` to length `num_envs - 1`. This happens when reconfiguration is enabled. Here is an example:

```python
import gymnasium as gym
import mani_skill.envs

env = gym.make("PickCube-v1", num_envs=4, obs_mode="state", sim_backend="gpu", reconfiguration_freq=1)

env.reset(seed=1)
base = env.unwrapped
print(base.num_envs)            # 4
print(len(base._episode_seed))  # actual: 4, expected: 4

env.reset()
print(len(base._episode_seed))  # actual: 3, expected: 4
```

The likely cause is `_set_main_rng` using `+` between a numpy array and a python list, while `_set_episode_rng` uses `np.concatenate`. This leaves `_batched_main_rng` with size `num_envs - 1`, and the later unseeded reset uses it to generate episode seeds. Using `np.concatenate` fixes this issue.

HaozheZhang6:
## What

`BaseEnv._set_main_rng` expanded a length-1 main seed with `array + list`:

```python
self._main_seed = self._main_seed + np.random.RandomState(...).randint(2**31, size=(self.num_envs - 1,)).tolist()
```

`np.array([seed]) + [a, b, c]` is a numpy **broadcast**, not a concatenation, so it yields `num_envs - 1` entries (`[seed+a, seed+b, seed+c]`) instead of `num_envs`. `_batched_main_rng` is then built from that short list, and a later unseeded reset under reconfiguration seeds the episodes from it (`_set_episode_rng(self._batched_main_rng.randint(...))`), leaving `_episode_seed` at `num_envs - 1`.

Repro from #1456 (`num_envs=4`):
```python
env.reset(seed=1); len(base._episode_seed)  # 4
env.reset();       len(base._episode_seed)  # 3  <- should be 4
```

## Fix

Use `np.concatenate`, matching the sibling `_set_episode_rng` which already does this:
```python
self._main_seed = np.concatenate((self._main_seed, np.random.RandomState(self._main_seed[0]).randint(2**31, size=(self.num_envs - 1,))))
```
This also corrects the seed *values* — the first env keeps the user seed instead of `seed + derived`.

## Test

Added `test_seeded_then_unseeded_reset_keeps_episode_seed_batch_size` to `tests/test_gpu_envs.py` (the bug needs `num_envs > 1`, i.e. the CUDA backend). Unlike the commented-out reproducibility tests there, it only asserts the `_episode_seed` batch size — deterministic even though GPU sim isn't — so it's safe to run.

I verified the broadcast-vs-concatenate behaviour locally in isolation; I don't have a CUDA box to run the env-level test, so it targets your GPU CI.

Fixes #1456

### DIFF
--- a/mani_skill/envs/sapien_env.py
+++ b/mani_skill/envs/sapien_env.py
@@ -1001,7 +1001,7 @@ class BaseEnv(gym.Env):
         self._main_seed = seed_list
         self._main_rng = np.random.RandomState(self._main_seed[0])
         if len(self._main_seed) == 1 and self.num_envs > 1:
-            self._main_seed = self._main_seed + np.random.RandomState(self._main_seed[0]).randint(2**31, size=(self.num_envs - 1,)).tolist()
+            self._main_seed = np.concatenate((self._main_seed, np.random.RandomState(self._main_seed[0]).randint(2**31, size=(self.num_envs - 1,))))
         self._batched_main_rng = BatchedRNG.from_seeds(self._main_seed, backend=self._batched_rng_backend)
 
     def _set_episode_rng(self, seed: Union[None, list[int], np.ndarray[Any], int], env_idx: torch.Tensor):
--- a/tests/test_gpu_envs.py
+++ b/tests/test_gpu_envs.py
@@ -189,6 +189,24 @@ def test_env_reconfiguration(env_id):
 #     del env
 
 
+# Unlike the reproducibility tests above, this only checks the batch size of the episode seeds, which is
+# deterministic even though GPU sim is not.
+@pytest.mark.gpu_sim
+def test_seeded_then_unseeded_reset_keeps_episode_seed_batch_size():
+    # Regression for #1456: a scalar-seeded reset expanded the main seed with `+` (numpy broadcast, which
+    # yields num_envs-1 entries) instead of np.concatenate, so a following unseeded reset under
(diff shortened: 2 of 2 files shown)


HaozheZhang6:
@StoneT2000 gentle nudge on this -- green for a week now. It fixes the scalar-seeded main RNG expansion that was shrinking per-env entropy. Let me know if you'd like any changes.

StoneT2000:
oh my, that seems very bad...

this will change a lot of RNG in maniskill so probably won't make it a new pypi package but it can go to nightly. I'll check tonight.

Will make sure it goes to the maniskill 4 release (where RNG is already not guaranteed to be the same anyway)

HaozheZhang6:
Makes sense -- nightly + the ManiSkill 4 release is the right home for it given it shifts the RNG stream. The change itself is tiny (a single `+` -> `np.concatenate`, matching what `_set_episode_rng` already does), so it should be quick to review tonight.

StoneT2000 (approved):

PR #1466, opened 2026-07-06
Viser integration

jstmn:
This PR adds a viser visualizer to maniskill. 

Todos:
1. Toggle to show the coordinate frame of every link
2. Toggle to show the joint axes of every joint
3. ?

https://github.com/user-attachments/assets/41bef753-d03a-4f8b-81e5-151e41153597


jstmn:
https://github.com/user-attachments/assets/1548e477-23e3-439f-b929-c8ec6694b3a4

Ready for review / merging

mani_skill/envs/tasks/tabletop/pick_cube.py
```diff
@@ -30,7 +30,7 @@
 """
 
 
-@register_env("PickCube-v1", max_episode_steps=50)
+@register_env("PickCube-v1", max_episode_steps=50000)
```
StoneT2000: remove this change

mani_skill/envs/utils/system/backend.py
StoneT2000, line 15: backend info is exclusively used for physics and rendering engine, not visualizer.

Visualizer itself is a bit of a special entity and should have it's own  handling

moreover as we do the newton integration, some of the files like sapien_env.py and backend.py are being refactored and changed quite a bit, I'd recommend trying to make new files instead of editing existing ones too much, will make it easier for us to port it to maniskill 4 when that is ready.

mani_skill/envs/sapien_env.py
```diff
@@ -742,11 +742,18 @@ def _reconfigure(self, options = dict()):
         self._load_agent(options)
 
         self._load_scene(options)
-        if self.scene.can_render(): self._load_lighting(options)
+        # viser has a real (CPU) render device so entities build normal visual components (see
+        # parse_sim_and_render_backend), but does not support SAPIEN's own lighting pipeline, so that is
+        # skipped here even though self.scene.can_render() is True
+        if self.scene.can_render() and not self.scene.viser_enabled:
+            self._load_lighting(options)
```
StoneT2000: why if viser is enabled we don't load default lighting?

mani_skill/envs/scene.py
```diff
@@ -33,6 +35,414 @@
     sapien.render.RenderCameraGroup = "oldtype"
 
 
+@dataclass
```
StoneT2000: none of the viser code should go into the scene.py file, make a separate file for it and import it into scene.py

mani_skill/utils/geometry/trimesh_utils.py
```diff
@@ -50,6 +53,15 @@ def get_render_body_meshes(visual_body: sapien.render.RenderBodyComponent):
...
+def _set_mesh_color(mesh: trimesh.Trimesh, material: sapien.render.RenderMaterial):
+    """Colors a mesh with a render material's base_color. Ignores textures for now, so
+    texture-mapped materials (e.g. wood grain) will just show up as their flat base_color."""
+    color = np.clip(np.array(material.base_color) * 255, 0, 255).astype(np.uint8)
+    mesh.visual = trimesh.visual.ColorVisuals(
+        mesh, face_colors=np.tile(color, (len(mesh.faces), 1))
+    )
+
+
 def get_render_shape_meshes(render_shape: sapien.render.RenderShape):
```
StoneT2000, line 65: why was this changed? it's not used by viser code

StoneT2000 (changes requested):

StoneT2000:
see my requested changes. After we split off your changes into new files instead of editing existing ones too much, i'll take a deeper look into the exact viser code on next review. Thanks!

mani_skill/envs/tasks/tabletop/pick_cube.py
jstmn: done

mani_skill/envs/utils/system/backend.py
jstmn: gotcha makes sense. reverted backend.py to version on main and added `visualizer_backend` that can either be 'sapien' or 'vizer' instead.

just curious what's going to get ported from newton? mjwarp maybe? rendering too?

mani_skill/envs/sapien_env.py
jstmn: removed

mani_skill/envs/scene.py
jstmn: moved to viser_visualizer.py

mani_skill/utils/geometry/trimesh_utils.py
jstmn: Without these changes the meshes are all monotone white! Viser uses trimesh under the hood for loading / using meshes. Currently the `mesh` variables don't have `.visual` set afaict. The change that's made is here:

```
    mesh.visual = trimesh.visual.ColorVisuals(
        mesh, face_colors=np.tile(color, (len(mesh.faces), 1))
    )
```

jstmn:
updated! I added `visualizer_backend` which can either be `sapien` or `viser`. I think that's a fairly clean way to go. It maintains the render_backend, sim_backend naming convention as well.

jstmn:
Screenshot 2026-07-11 at 10 22 31 AM

Added AABB, bounding box, 3d point list, coordinate frames to match the same API in sapien: https://github.com/haosulab/SAPIEN/blob/master/python/py_package/utils/viewer/viewer.py#L299C1-L466C44

PR #1467, opened 2026-07-06
fix(act): handle CPU eval metrics without final_info

### LINKED ISSUE #1408: Evaluate.py for ACT policy does not work
When running the baseline.sh script for ACT, a Key error raises when evaluating the environment.

Traceback (most recent call last):
  File "/home/lzanetti/workspaces/latent_vla_ws/src/ManiSkill/examples/baselines/act/train.py", line 428, in <module>
    eval_metrics = evaluate(args.num_eval_episodes, ema_agent, envs, eval_kwargs)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lzanetti/workspaces/latent_vla_ws/src/ManiSkill/examples/baselines/act/act/evaluate.py", line 83, in evaluate
    if isinstance(info["final_info"], dict):
                  ~~~~^^^^^^^^^^^^^^
KeyError: 'final_info'

This is because when running train.py, using "physx_cpu", the CPUGymWrapper is used and that does not include "final_info" in the info dict when truncation/termination happens

saime428:
## Summary
- Handle ACT evaluation metrics from CPU vector envs that expose metrics under `info["episode"]` instead of `info["final_info"]`.
- Preserve the existing `final_info` handling for ManiSkillVectorEnv/GPU paths.
- Normalize tensor metrics to CPU NumPy arrays across supported info shapes and exclude Gymnasium `_`-prefixed mask keys from returned metrics.
- Add focused regression coverage for the supported ACT evaluation info shapes.

## Why
- Fixes `KeyError: 'final_info'` when ACT evaluation runs with `physx_cpu` and `CPUGymWrapper(record_metrics=True)`.
- Keeps metric element types consistent before the existing `np.stack` step.
- Prevents Gymnasium mask arrays from being returned as metrics.
- Closes #1408.

## Validation
- [x] `python -m compileall examples/baselines/act/act/evaluate.py`
- [x] `pytest tests/test_act_evaluate.py -q`

## Notes for reviewers
- Full ManiSkill simulation smoke test was not run locally because `sapien` is not installed in this environment.
- Handling `None` entries inside asynchronous `final_info` arrays remains unchanged; this evaluation loop currently asserts synchronous truncation.

### DIFF
--- a/examples/baselines/act/act/evaluate.py
+++ b/examples/baselines/act/act/evaluate.py
@@ -21,8 +21,12 @@ def _append_eval_metrics(eval_metrics, info):
         if isinstance(final_info, dict):
             _append_episode_metrics(eval_metrics, final_info["episode"])
         else:
-            for final_info in final_info:
-                _append_episode_metrics(eval_metrics, final_info["episode"])
+            # Gymnasium leaves None in final_info for envs that did not terminate, and the
+            # companion _final_info mask says which entries are real. Neither is checked
+            # here because the caller asserts every env truncates on the same step, so this
+            # branch only ever sees a fully populated array.
+            for env_info in final_info:
+                _append_episode_metrics(eval_metrics, env_info["episode"])
         return
     if "episode" in info:
         _append_episode_metrics(eval_metrics, info["episode"])
--- a/examples/baselines/act/act/evaluate.py
+++ b/examples/baselines/act/act/evaluate.py
@@ -5,6 +5,31 @@ import torch
 
 from mani_skill.utils import common
 
+
+def _append_episode_metrics(eval_metrics, episode):
+    for k, v in episode.items():
+        if k.startswith("_"):
+            continue
+        if isinstance(v, torch.Tensor):
+            v = v.float().cpu().numpy()
+        eval_metrics[k].append(v)
+
+
+def _append_eval_metrics(eval_metrics, info):
+    if "final_info" in info:
+        final_info = info["final_info"]
+        if isinstance(final_info, dict):
+            _append_episode_metrics(eval_metrics, final_info["episode"])
+        else:
+            for final_info in final_info:
+                _append_episode_metrics(eval_metrics, final_info["episode"])
+        return
+    if "episode" in info:
+        _append_episode_metrics(eval_metrics, info["episode"])
+        return
+    raise KeyError("expected info['final_info'] or info['episode'] with episode metrics")
+
+
 def evaluate(n: int, agent, eval_envs, eval_kwargs):
     stats, num_queries, temporal_agg, max_timesteps, device, sim_backend = eval_kwargs.values()
 
@@ -80,13 +105,7 @@ def evaluate(n: int, agent, eval_envs, eval_kwargs):
             # collect episode info
             if truncated.any():
                 assert truncated.all() == truncated.any(), "all episodes should truncate at the same time for fair evaluation with other algorithms"
-                if isinstance(info["final_info"], dict):
-                    for k, v in info["final_info"]["episode"].items():
-                        eval_metrics[k].append(v.float().cpu().numpy())
-                else:
-                    for final_info in info["final_info"]:
-                        for k, v in final_info["episode"].items():
-                            eval_metrics[k].append(v)
+                _append_eval_metrics(eval_metrics, info)
                 # new episodes begin
                 eps_count += num_envs
                 ts = 0
--- a/tests/test_act_evaluate.py
+++ b/tests/test_act_evaluate.py
@@ -1,0 +1,97 @@
+import importlib.util
+import sys
+import types
+from collections import defaultdict
+from pathlib import Path
+
+import numpy as np
+import pytest
+import torch
+
+
+def load_act_evaluate():
+    common = types.ModuleType("common")
+    utils = types.ModuleType("mani_skill.utils")
+    utils.common = common
+
+    sys.modules["mani_skill"] = types.ModuleType("mani_skill")
+    sys.modules["mani_skill.utils"] = utils
+    sys.modules["mani_skill.utils.common"] = common
+
+    evaluate_path = (
+        Path(__file__).resolve().parents[1]
+        / "examples"
+        / "baselines"
+        / "act"
+        / "act"
+        / "evaluate.py"
+    )
+    spec = importlib.util.spec_from_file_location("act_evaluate", evaluate_path)
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+def test_append_eval_metrics_reads_vector_final_info_dict():
+    module = load_act_evaluate()
+    metrics = defaultdict(list)
+
+    module._append_eval_metrics(
+        metrics,
+        {
+            "final_info": {
+                "episode": {
+                    "return": torch.tensor([1.0, 2.0]),
+                    "success_at_end": torch.tensor([True, False]),
+                }
+            }
+        },
+    )
+
+    np.testing.assert_array_equal(metrics["return"][0], np.array([1.0, 2.0]))
+    np.testing.assert_array_equal(metrics["success_at_end"][0], np.array([1.0, 0.0]))
+
+
+def test_append_eval_metrics_reads_vector_final_info_list():
+    module = load_act_evaluate()
+    metrics = defaultdict(list)
+
+    module._append_eval_metrics(
+        metrics,
+        {
+            "final_info": [
+                {"episode": {"return": np.array(1.0)}},
+                {"episode": {"return": np.array(2.0)}},
+            ]
+        },
+    )
+
+    assert metrics["return"] == [np.array(1.0), np.array(2.0)]
+
+
+def test_append_eval_metrics_falls_back_to_cpu_episode_info():
+    module = load_act_evaluate()
+    metrics = defaultdict(list)
+
+    module._append_eval_metrics(
+        metrics,
+        {
+            "episode": {
+                "return": np.array([1.0, 2.0]),
+                "_return": np.array([True, True]),
+                "success_at_end": np.array([True, False]),
+                "_success_at_end": np.array([True, True]),
+            }
+        },
+    )
+
+    assert set(metrics) == {"return", "success_at_end"}
+    np.testing.assert_array_equal(metrics["return"][0], np.array([1.0, 2.0]))
+    np.testing.assert_array_equal(metrics["success_at_end"][0], np.array([True, False]))
+
+
+def test_append_eval_metrics_requires_episode_metrics():
+    module = load_act_evaluate()
+
+    with pytest.raises(KeyError, match="final_info.*episode"):
+        module._append_eval_metrics(defaultdict(list), {})


saime428:
Hi, checking in on this fix for #1408. The change is limited to handling the CPU vector-env info["episode"] shape while preserving the existing final_info path, with focused tests. Please let me know if you'd like any changes or additional validation.

ErenAta16 (commented):
The stated fix works — I ran the old and new helpers side by side against the four info shapes:

```
{"episode": {...}}                     old: KeyError: 'final_info'   new: collected
{"final_info": {"episode": {...}}}     both collected
{"final_info": [{"episode": {...}}]}   both collected
```

Extracting the dispatch into `_append_eval_metrics` is also the right call — the old inline version had the two branches drift apart, which is the second point below.

Two behaviour changes come along with it that the description doesn't mention. Both look like improvements to me, but they change what `evaluate()` returns, and the consumer is:

```python
for k in eval_metrics.keys():
    eval_metrics[k] = np.stack(eval_metrics[k])
return eval_metrics
```

so anything that changes the keys or the element types changes the returned dict.

## The two branches converted differently, and now they don't

On `main`:

```python
if isinstance(info["final_info"], dict):
    eval_metrics[k].append(v.float().cpu().numpy())   # converts
else:
    eval_metrics[k].append(v)                          # does not
```

Measured element types:

```
list branch, tensor values   old: Tensor    new: ndarray
dict branch, tensor values   old: ndarray   new: ndarray
```

Unifying them is clearly right — and it matters more than cosmetics, because `np.stack` on a list of CUDA tensors raises, so the list branch on `main` could only ever have worked with values that were already numpy or on CPU. The new conversion removes that dependency. Worth a line in the description: it's a second fix, not a refactor.

## `_`-prefixed keys are now dropped

```python
if k.startswith("_"):
    continue
```

is new, and it applies to both branches:

```
{"success": tensor, "_success": ndarray}
  old -> keys: success, _success
  new -> keys: success
```

Skipping gymnasium's `_`-prefixed mask arrays is almost certainly what you want — they were being stacked and returned as if they were metrics. But callers reading `evaluate()`'s result by key, or logging every key, will see the set shrink. Cheap to mention in the PR body; harder to explain later when someone's dashboard loses a series.

## One thing neither version handles

`for final_info in final_info:` walks the array without checking for `None`, and gymnasium's vector envs put `None` in `final_info` for envs that didn't terminate:

```
{"final_info": np.array([{...}, None], dtype=object)}
  old -> TypeError: 'NoneType' object is not subscriptable
  new -> TypeError: 'NoneType' object is not subscriptable
```

Pre-existing, not introduced here, and the `assert truncated.all() == truncated.any()` above means every env terminates together in this loop — so it may be unreachable by construction. Mentioning it because the `k.startswith("_")` filter shows awareness of the `_final_info` mask convention, and the mask exists precisely to say which entries are real. If the assert is the reason it's safe, a comment saying so would stop the next person from adding a redundant guard.

## Smaller

`for final_info in final_info:` rebinds the name being iterated. It works, since the iterator is created before the first assignment, but it makes the loop harder to read than `for env_info in final_info:` and would silently change meaning if anyone later moved the loop or added a second pass over the same variable.

saime428:
Thanks for the detailed validation. I updated the PR description to call out both behavior changes: tensor metrics are normalized to CPU NumPy arrays, and Gymnasium `_`-prefixed mask keys are excluded. I left the pre-existing `None` entry behavior unchanged because this PR stays scoped to the current synchronous-truncation evaluation path.

saime428:
Picking up the two points I left on the table in 134eb5a — sorry for the delay, my earlier reply only covered the description changes.

**The rebound loop variable.** Renamed to `env_info`. You were right that it was worth doing rather than leaving: it worked only because the iterator is created before the first assignment, which is a property nobody should have to re-derive when they touch this loop.

**Why `None` cannot appear here.** Your read is correct — the `assert truncated.all() == truncated.any()` in the caller is the reason, and that reasoning was living in this thread instead of in the code. It is now a comment at the branch:

```python
else:
    # Gymnasium leaves None in final_info for envs that did not terminate, and the
    # companion _final_info mask says which entries are real. Neither is checked
    # here because the caller asserts every env truncates on the same step, so this
    # branch only ever sees a fully populated array.
    for env_info in final_info:
        _append_episode_metrics(eval_metrics, env_info["episode"])
```

I still think adding a runtime guard belongs in whatever change makes truncation non-simultaneous, not this one — but the next person should not have to reconstruct that argument, which was your point.

Re-ran the shapes you listed against the edited file to confirm the rename is behaviour-preserving:

```
{"episode": {...}}                     -> success: [ndarray]
{"final_info": {"episode": {...}}}     -> success: [ndarray]
{"final_info": [{...}, {...}]}         -> success: [ndarray, ndarray]
{"nothing": 1}                         -> KeyError
```

The two-element case is the one that matters for the rename — it confirms the loop still visits every entry rather than only the last. `_`-prefixed keys stay excluded. The PR's own tests are 4 passed.

Thanks for the review — the side-by-side runs of the old and new helpers were more than I expected and made both points easy to act on.

ErenAta16 (approved):
Both points are addressed and I re-ran the shapes against the edited file rather than taking the summary.

```
info shape                           result
{"episode": {...}}                   success: [ndarray]
{"final_info": {"episode": {...}}}   success: [ndarray]
{"final_info": [{...}]}              success: [ndarray]
{"final_info": [{...}, {...}]}       success: [ndarray, ndarray]
{"nothing": 1}                       KeyError("expected info['final_info'] or info['episode'] ...")
```

The two-element row is the one that matters for the rename, and it visits both entries, so `env_info` is behaviour-preserving. `_`-prefixed keys stay excluded on every path.

The drift I raised is gone as a structural property rather than by matching the two branches by hand:

```
dict branch : success: [ndarray]
list branch : success: [ndarray]
```

Both now route through `_append_episode_metrics`, so the `v.float().cpu().numpy()` on one side and the bare `v` on the other cannot come apart again. That is worth more than the original fix, because the old shape would have let the next edit reintroduce the mismatch silently.

**On the runtime guard, I agree with you and I am withdrawing the suggestion.** The comment is the right artifact here:

> Neither is checked here because the caller asserts every env truncates on the same step, so this branch only ever sees a fully populated array.

An assertion in this function would be checking something the caller already guarantees, and it would go stale the moment that guarantee moves. What was actually missing was the reason, and the reason was sitting in a review thread where nobody would find it. Adding the guard belongs to whatever change makes truncation non-simultaneous, exactly as you say, and that change now has a comment pointing at what it needs to revisit.

Approving. Thanks for going back for the two loose ends rather than letting them ride, and for re-running the cases instead of asserting the rename was safe.

saime428:
Appreciate you re-running the shapes against the edited file instead of taking my summary for it. The branch drift you caught on the first pass is what turned this from a one-line KeyError fix into something structural — left to myself I'd have shipped the narrow version and let the two conversions come apart again a few edits later.

The framing on the guard stuck with me too. What was missing wasn't a check, it was the reason, and the reason was sitting in a review thread where nobody would go looking. That's a better way to think about it than my "out of scope for this PR" answer was.

ErenAta16:
That is generous, and I want to give the better half of it back to you.

The branch drift was only visible because you had written the PR so the two conversion paths sat next to each other in the diff. Plenty of fixes for this class of bug hide the second path in a different file, and then nobody notices they have come apart until someone bisects a metrics regression six months later. That was your structure, not my reading.

The part I got the most out of was your reply on the guard. I framed it as "add a runtime check or explain why you cannot", and your answer was better: what was missing was not a check, it was the reason, and the reason was sitting in a review thread where nobody would ever look for it. Putting it in a comment at the branch means the next person gets the argument without having to reconstruct it or come find us. I have used that framing twice since.

Also worth saying plainly: renaming the rebound loop variable when it already worked, and re-running the four shapes against the edited file rather than telling me it was fine, is not the common response to review feedback. It made this an easy PR to approve.

Hope a maintainer picks it up soon. If it goes stale and you would like a second voice on the thread, tag me.

saime428:
Thanks for the really thoughtful review and approval — your feedback made this a lot more solid than the original narrow fix.

The loop variable rename is a good example. I had convinced myself it was fine, but you're right that leaving it as "it works because the iterator is created first" is just debt waiting to be rediscovered. Making it explicit cost nothing.

The guard comment has already been useful — I ended up using the same "explain why we can skip the check" framing in another PR.

Really appreciate you running the old and new helpers side by side. That comparison is what caught the branch drift, and turned what could have been a silent future breakage into something more robust.

Hoping a maintainer picks this up soon. If it goes stale I'll definitely tag you.

saime428:
@StoneT2000 this one is approved and idle — could you take a look when you have a moment?

Short version: `evaluate` in the ACT example crashed with `KeyError` whenever the CPU env returned an `info` without `final_info`. @ErenAta16 reviewed it twice, ran all four `info` shapes against the edited file both times, and approved on 2026-08-07. Nothing has changed on the branch since.

Two behavior changes are worth your eyes specifically, since they go slightly past the reported `KeyError`:

- tensor metrics are normalized to CPU NumPy arrays
- Gymnasium `_`-prefixed mask keys are excluded

Both are called out in the PR description. Pre-existing `None`-entry behavior is left as it was. Happy to change anything you would rather see done differently.

saime428:
Closing this to keep the queue clear.

Thanks @ErenAta16 for the two rounds of review, and for running the actual info shapes against the helper rather than reading the diff. That was more careful than the change deserved.

The branch stays up if this is ever useful.

ErenAta16:
Thanks for saying so, and for the two rounds of turning the review around quickly
enough that running the shapes was actually worth doing. That only pays off when
the author is still around to act on it, which is not a given on a branch this old.

One thing worth leaving on the record now that it is closed: the bug has not gone
anywhere. `examples/baselines/act/act/evaluate.py` on `main` still reads

```python
if isinstance(info["final_info"], dict):
    for k, v in info["final_info"]["episode"].items():
...
    for final_info in info["final_info"]:
        for k, v in final_info["episode"].items():
```

at lines 83 to 88, unguarded, which is exactly the shape that raises `KeyError`
on the CPU path. So closing this cleared your queue rather than the defect, and
your branch is still the fix if anyone picks it up. Keeping it up was the right
call.

If you do come back to it, the thing I would change is nothing in the code and
one line in the description: lead with the four info shapes and which ones raise
today. A reviewer who sees `KeyError` next to a concrete input list decides in a
minute; one who has to reconstruct when `final_info` is absent puts it back in the
queue. That is most of the difference between two months of silence and a merge,
and it is not really about the patch.

saime428:
Checked the lines you quoted before taking your word for it: `evaluate.py:83-88` on `main` is unguarded exactly as you describe, so you are right that closing this cleared my queue and not the defect.

The description point landed harder. I put this PR next to one of mine that merged in a day: this one opens with four lines of what the patch does and puts `KeyError: 'final_info'` in the second section, and the one that merged opens with the failing behaviour. Same care either way, very different first ten seconds for whoever picks it up. I am rewriting another open PR's description on that basis, so the advice is already doing something.

Thanks for spending the time on a branch that was already closed.

ErenAta16:
Checking the lines rather than taking my word for it is the part worth noting back at you. Plenty of people would have accepted a quoted snippet from a stranger, and the one time that snippet is stale you end up defending a claim that was never yours.

The comparison you ran is a better version of the advice than the advice was. I told you to lead with the failing behaviour; you went and measured it against a PR of your own that merged in a day and found the difference was the first ten seconds. That is a more useful thing to carry to the next one than anything I said, because it came with a control.

If you want a second pair of eyes on a description before you post it, send me the link and I will read it as a reviewer seeing it cold, which is the only angle that actually tests whether the opening works. No obligation either way.

PR #1470, opened 2026-07-17
[BugFix] Fix flatten_state_dict when use_torch=False

ywchoi02:
## Summary

This fixes an issue in `flatten_state_dict` when `use_torch=False`.

Previously, recursive calls always checked `state.nelement()`, which raises an `AttributeError` for NumPy arrays because `numpy.ndarray` does not implement `nelement()`.

This PR:
- uses `state.size` when `use_torch=False`;
- passes `device` through recursive calls to preserve the requested device.

## Example

```python
# Before:
flatten_state_dict(state_dict, use_torch=False)
# AttributeError: 'numpy.ndarray' object has no attribute 'nelement'

# After:
flatten_state_dict(state_dict, use_torch=False)
# Works as expected.

### DIFF
--- a/mani_skill/utils/common.py
+++ b/mani_skill/utils/common.py
@@ -230,11 +230,13 @@ def flatten_state_dict(
 
     for key, value in state_dict.items():
         if isinstance(value, dict):
-            state = flatten_state_dict(value, use_torch=use_torch)
-            if state.nelement() == 0:
+            state = flatten_state_dict(
+                value,
+                use_torch=use_torch,
+                device=device,
+            )
+            if (state.nelement() if use_torch else state.size) == 0:
                 state = None
-            elif use_torch:
-                state = to_tensor(state, device=device)
         elif isinstance(value, (tuple, list)):
             state = None if len(value) == 0 else value
             if use_torch:

### CODE REVIEW

**Review by ErenAta16 — COMMENTED** [review]
Reproduced both revisions of `flatten_state_dict` side by side rather than reading the diff, since the change is small enough that the interesting part is what it does and doesn't cover.

## The `AttributeError` is real and the fix is complete

Any nested dict on the numpy path hits it, not just the example in the description:

| input | `main` | this PR |
|---|---|---|
| `{"a": arr, "b": {"c": arr}}`, `use_torch=False` | `AttributeError: 'numpy.ndarray' object has no attribute 'nelement'` | `array([1., 2., 3.])` |
| `{"a": arr, "b": {}}`, `use_torch=False` | same `AttributeError` | `array([1.])` |
| `{"a": {"b": {"c": arr}}}`, `use_torch=False` | same `AttributeError` | `array([1.])` |
| flat dict, `use_torch=False` | `array([1., 2., 3.5])` | unchanged |
| any of the above, `use_torch=True` | correct | unchanged |

So it isn't only an empty-dict edge case — `use_torch=False` is unusable with *any* nesting on `main`, and the fix restores it without touching the torch path.

## The device half of the description is stronger than what changed

> passes `device` through recursive calls to preserve the requested device

The device was already preserved on `main`, by the line the PR removes:

```python
state = flatten_state_dict(value, use_torch=use_torch)
if state.nelement() == 0:
    state = None
elif use_torch:
    state = to_tensor(state, device=device)   # <- moved it here
```

Checked with a `meta` device so the placement is observable:

```
main -> meta
PR   -> meta
```

What actually changes is the number of transfers. Counting `to_tensor` calls on `{"a": arr, "b": {"c": arr, "d": arr}, "e": {"f": arr}}` with `use_torch=True`:

```
main : 6
PR   : 4
```

Nested tensors are now allocated on the target device instead of being built on the default device and moved afterwards. That's a real improvement and worth keeping — but it's an allocation saving, not a device-correctness fix, and describing it as the latter sends the next reader looking for a bug that isn't there. Worth rewording so the `AttributeError` stays the headline.

## Two things one branch away, both pre-existing

Neither is this PR's job, but they're the same defect class in the same function, so flagging them here rather than opening separate issues:

**A raw `torch.Tensor` value never honours `device`.**

```python
elif isinstance(value, torch.Tensor):
    state = value
    if len(state.shape) == 1:
        state = state[:, None]
```

No `to_tensor`, so a CPU tensor in the dict is hstacked with siblings built on the requested device:

```
{"t": torch.tensor([1., 2.]), "a": np.array([3.])}, use_torch=True, device=meta
-> RuntimeError: Tensor on device meta is not on the expected device cpu!
```

Same on `main` and on this branch. Every other branch guards with `if use_torch: state = to_tensor(state, device=device)`; this one doesn't.

**The same input shape flattens differently depending on its type.**

```
{"a": np.array([1., 2.])}      -> [1., 2.]
{"t": torch.tensor([1., 2.])}  -> [[1.], [2.]]
```

The `state[:, None]` reshape applies only to the tensor branch, and it applies on the numpy path too — `use_torch=False` with a tensor value returns an ndarray of shape `(2, 1)` where the equivalent ndarray input gives `(2,)`. A caller mixing the two in one state dict gets a silently different flattened width.

Happy to open these as an issue if you'd rather keep this PR to the one line.

ErenAta16:
Filed the two adjacent items as #1478 rather than widening this PR, since the dimension one needs a decision on whether the function's batching contract is per-env or not.

Worth knowing while reviewing this: the mixed-type failure is not specific to the `use_torch=False` path this PR fixes. It reproduces on both:

```
{"a": np.array([1., 2.]), "t": torch.tensor([3., 4.])}
  use_torch=False -> ValueError: all the input arrays must have same number of dimensions
  use_torch=True  -> RuntimeError: Tensors must have same number of dimensions: got 2 and 1
```

So it's independent of this change and won't be masked by it — this PR is still the right fix for the `AttributeError`.

Issue #1435, opened 2026-06-09
EE_Pose and EE_Delta_Pose issues with Unitree G1


Lrogic:
## Problem

I'm trying to control the right hand of the Unitree G1 robot using the **PDEEPoseControllerConfig** setup. I extended the UnitreeG1UpperBody subclass with my own implementation. However, the issue is when I try to control the right hand with `ee_pose`, the robot moves very slightly then freezes (videos below). The same thing with `ee_delta_pose`. I'm hoping someone can help me debug these issues.

## Setup

**Waypoint sequence**
The right hand is commanded to try to move to three goal poses defined by a 7D vector: [x, y, z, qw, qx, qy, qz]. A waypoint is considered achieved when the hand is within `pos_tol` distance and `rot_tol` rotation distance (described below).

```javascript
{
  "poses": [
    [-0.08, 0.24, 0.845, 0.7071068, 0.0, 0.0, -0.7071068],
    [-0.15, 0.18, 0.80, 0.7071068, 0.0, 0.0, -0.7071068],
    [-0.20, 0.025, 0.6409, 0.7068956888241477, 0.707317701654041, 0.00018099998471816394, 0.00034826381457648224]
  ],
  "pos_tol": 0.03,
  "rot_tol": 0.5,
  "stable_steps": 10
}
``` 
**Control loop**
Each simulation step, I compute a right_arm action (delta or absolute, depending on control mode) that moves the right TCP toward the current waypoint in world frame. The controller advances to the next waypoint only when the right TCP stays within pos_tol and rot_tol for stable_steps consecutive steps (10 in the config above). Only the right_arm action slice is set; right-hand and passive joint actions are zero.

## Failure videos

Here are the videos of the robot trying to move its right hand to several goal waypoints which are displayed as green cubes in the video. Once it achieves a waypoint, it sets the next target location as the next waypoint.

https://github.com/user-attachments/assets/990f3d56-0679-4b2f-a373-3c85bff1a939
**ee_delta_pose**: It achieves the first waypoint, then gets stuck after that, when the next waypoint is set as the goal.

https://github.com/user-attachments/assets/69b2ee3f-1948-4b11-a832-51f611f5c148
**ee_pose**: It achieves the first waypoint, then gets stuck after that, when the next waypoint is set as the goal.

## Expected vs Actual
| Mode | Expected | Actual |
|------|----------|--------|
| `pd_ee_delta_pose` | Reach all 3 waypoints | Reaches first, then stuck |
[...]

PR #1472, opened 2026-07-24
[BugFix] scale PDEEPoseController rotation by rot_upper so action sign is preserved

### LINKED ISSUE #1469: [Bug] PDEEPoseController scales rotation actions by rot_lower (negative), inverting commanded rotation direction
## Summary

`PDEEPoseController._clip_and_scale_action` scales the (norm-clipped) rotation action by `self.config.rot_lower`:

```python
rot_action = rot_action * self.config.rot_lower
```

`rot_lower` is negative (e.g. `-0.1` in the Panda/XArm6 `pd_ee_delta_pose` configs; the `PDEEPoseControllerConfig` default on current `main` is `-2 * np.pi`), so **a positive normalized rotation action produces a negative realized rotation** — the commanded rotation direction is inverted. Translation actions go through `clip_and_scale_action(low, high)` and behave as expected; only the rotation path has this asymmetry.

## Reproduction (mani_skill 3.0.1, CPU sim, `pd_ee_delta_pose`, Panda)

```python
import gymnasium as gym, numpy as np, mani_skill.envs

env = gym.make("PickCube-v1", num_envs=1, obs_mode="state",
               control_mode="pd_ee_delta_pose", sim_backend="physx_cpu")
env.reset(seed=0)
arm = env.unwrapped.agent.controller.controllers["arm"]

a = np.zeros(7, dtype=np.float32)
a[5] = 1.0            # positive rotation action about z, normalized
env.step(a)
print(arm._target_pose) # target orientation is rotated NEGATIVELY about z
```

Measured on a state-mode tabletop env with the stock Panda controller config (rot bounds ±0.1):

| normalized action `a[5]` | controller `_target_pose` rotation delta (1 step) | achieved EE rotation about z after 5 steps |
|---|---|---|
| `+1.0` | **−0.100 rad** | −0.196 rad |
| `−1.0` | **+0.100 rad** | +0.196 rad |

(The one-step target delta equals `rot_lower` exactly.)

## Why it hurts

Closed-loop Cartesian servoing that feeds back a standard rotation error (`rotvec of R_target · R_current⁻¹`) *diverges*: each correction increases the error. Because `compute_ik` failures make the controller silently hold the previous target (`_target_qpos = _start_qpos`), the wound-up wrist eventually presents as a frozen arm — the symptom looks like an IK bug rather than a sign inversion, which makes this expensive to debug.

## Expected

A positive normalized rotation action about +z should rotate the EE positively about +z (matching the translation semantics), i.e. scale by `rot_upper` or map through the same clip-and-scale as translation.

## Workaround

Negate the rotation part of the action before stepping.

Environment: mani_skill 3.0.1 (pip), sapien 3.0.3, CPU sim (`physx_cpu`), macOS arm64 / Python 3.12; the scaling code is unchanged on current `main`.

VihaanAgarwal:
Fixes #1469

## Problem

`PDEEPoseController._clip_and_scale_action` scales the norm-clipped rotation action by `self.config.rot_lower`:

```python
rot_action = rot_action * self.config.rot_lower
```

`rot_lower` is negative (`-2 * np.pi` by default, `-0.1` in the shipped Panda/XArm6 `pd_ee_delta_pose` configs), so a positive normalized rotation action produces a negative realized rotation. Translation goes through `clip_and_scale_action(low, high)` and keeps its sign, so only the rotation path was inverted.

This is easy to miss because `compute_ik` failures make the controller hold the previous target, so the wound-up wrist looks like a frozen arm rather than a sign error. Closed-loop Cartesian servoing that feeds back a standard rotation error diverges instead of converging.

## Fix

Scale the rotation by `rot_upper` (the positive bound), so a positive action rotates positively and matches the translation semantics. For the default symmetric bounds this only corrects the sign; the magnitude is unchanged.

## Tests

Added `tests/test_pd_ee_pose_controller.py`. These call `_clip_and_scale_action` directly, so they need no GPU or render backend (not marked `gpu_sim`) and run under `pytest tests/ -m "not slow and not gpu_sim"`:

- a positive rotation action about each axis stays positive, a negative one stays negative
- a full rotation action scales to the configured bound
- positive translation and positive rotation come out with the same (positive) direction

They fail on `main` with the inverted sign and pass with this change. `black`, `isort` and `autoflake` are clean on the touched files.

### DIFF
--- a/mani_skill/agents/controllers/pd_ee_pose.py
+++ b/mani_skill/agents/controllers/pd_ee_pose.py
@@ -245,7 +245,7 @@ class PDEEPoseController(PDEEPosController):
         rot_action[rot_norm > 1] = torch.mul(rot_action, 1 / rot_norm[:, None])[
             rot_norm > 1
         ]
-        rot_action = rot_action * self.config.rot_lower
+        rot_action = rot_action * self.config.rot_upper
         return torch.hstack([pos_action, rot_action])
 
     def compute_target_pose(self, prev_ee_pose_at_base: Pose, action):
--- a/tests/test_pd_ee_pose_controller.py
+++ b/tests/test_pd_ee_pose_controller.py
@@ -1,0 +1,76 @@
+"""Unit tests for PDEEPoseController action scaling.
+
+These exercise ``_clip_and_scale_action`` directly so they do not need a GPU or a
+render backend. See https://github.com/haosulab/ManiSkill/issues/1469 for the
+rotation sign bug this guards against.
+"""
+import pytest
+import torch
+
+from mani_skill.agents.controllers.pd_ee_pose import (
+    PDEEPoseController,
+    PDEEPoseControllerConfig,
+)
+
+
+def _make_scaler(rot_lower, rot_upper, pos_lower=-0.1, pos_upper=0.1):
+    """A stand-in exposing only what ``_clip_and_scale_action`` reads."""
+    config = PDEEPoseControllerConfig(
+        joint_names=[],
+        stiffness=0,
+        damping=0,
+        ee_link="ee",
+        urdf_path="",
+        pos_lower=pos_lower,
+        pos_upper=pos_upper,
+        rot_lower=rot_lower,
+        rot_upper=rot_upper,
+        frame="root_translation:root_aligned_body_rotation",
+    )
+    scaler = PDEEPoseController.__new__(PDEEPoseController)
+    scaler.config = config
+    scaler.action_space_low = torch.tensor(
+        [pos_lower] * 3 + [rot_lower] * 3, dtype=torch.float32
+    )
+    scaler.action_space_high = torch.tensor(
+        [pos_upper] * 3 + [rot_upper] * 3, dtype=torch.float32
+    )
+    return scaler
+
+
+@pytest.mark.parametrize("axis", [0, 1, 2])
+def test_rotation_action_preserves_sign(axis):
+    scaler = _make_scaler(rot_lower=-0.5, rot_upper=0.5)
+    action = torch.zeros((1, 6), dtype=torch.float32)
+    action[0, 3 + axis] = 1.0
+    scaled = scaler._clip_and_scale_action(action)
+    # a positive normalized rotation about an axis must stay positive, matching
+    # the translation semantics
+    assert scaled[0, 3 + axis].item() > 0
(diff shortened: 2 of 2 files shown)


StoneT2000:
Thank you for the possible fix. I actually think in maniskill 4 we might want to completely redo how PD EE Pose control works (the current way may be good, but it looks overly complicated). I am also a bit busy with MS4 recently and cannot properly review this until later, sorry!

abtonmoy:
Independent confirmation of this fix. I hit the same bug through #1469 and traced it before finding this PR, so for whenever review bandwidth frees up, here is corroborating evidence.

Root cause confirmed at `mani_skill/agents/controllers/pd_ee_pose.py` line 248 (main @ 62ff3a5): translation goes through `gym_utils.clip_and_scale_action(low, high)` and is correct, but the norm-clipped rotation action is multiplied by `self.config.rot_lower`, which is negative in the default config (`-2 * np.pi`) and in every shipped robot config. A normalized +1 rotation action therefore realizes exactly `rot_lower` radians, the opposite direction of the same action on the translation dims. Applies to delta and absolute variants, CPU and GPU, whenever `normalize_action=True` (the default). Scaling by `rot_upper`, as this PR does, restores the documented semantics; for the symmetric bounds used across the repo it changes only the sign.

Two things that may help this land:

1. I wrote regression tests that call the production `_clip_and_scale_action` directly with only the config and bound tensors, so they need no simulation backend, no GPU and no downloads. Five tests: positive rotation about each axis stays positive and reaches the bound, rotation and translation agree in direction for the same action, and norm clipping still applies. All five fail on current main and pass with this PR's change. Happy to contribute them to this PR or as a follow-up, whichever is easier.

2. One migration note worth a line in the changelog: any downstream code or trained policy that compensated by negating rotation actions will see flipped behavior after the fix. Demo datasets recorded through the controller are internally consistent either way; only callers that hardcoded the inversion are affected.

I also believe this bug explains the rotation-specific failures in #1435; analysis posted there.

VihaanAgarwal:
Appreciate the independent trace, that's exactly the root cause I hit: translation goes through clip_and_scale_action and is fine, but rotation gets multiplied by rot_lower, and the negative default flips the sign of every rotation command. Good to have #1469 as a second repro on record. No rush on review, this can sit until MS4 work settles, and I'm happy to rebase if the controller code moves in the meantime.

